PR #344, opened 2017-04-05
DevKnox pricing

dhilipsiva:


config/environment.js
```diff
@@ -4,6 +4,7 @@ module.exports = function(environment) {
   var ENV = {
     isDevknox: false,
     isAppknox: false,
+    devknoxPrice: 9,  // This should also change in `mycroft/settings.py`
```
subho007: @dhilipsiva why do we hardcode 9 here?

dhilipsiva: @subho007 For time being. We will have a config endpoint. There is an Issue there.

subho007: @dhilipsiva add a FIXME there. I'll approve the changes

dhilipsiva: @subho007 Done!

PR #567, opened 2018-01-05
Adds docker support for whitelabel build

Praseetha-KR:
- Modifies ember server to get `host` & `socketPath` from env if environment = `whitelabel`
- Docker entrypoint default command will:
   - build for environment = `whitelabel`
   - copies `dist` folder to nginx html folder 
   - start nginx in the foreground
- Show company logo in dashboard for whitelabel

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+node_modules/
+scripts/
+tests/
+tmp/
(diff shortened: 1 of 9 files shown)


yashwin:
@subho007 : test cases were failing here 😞

PR #665, opened 2018-08-08
Security Fixes & Enhancements 

yashwin:


app/models/security/analysis.js
```diff
@@ -17,7 +17,7 @@ export default DS.Model.extend({
   cvssVector: DS.attr('string'),
   cvssVersion: DS.attr('string'),
   analiserVersion: DS.attr('number'),
-  overriddenRisk: DS.attr('number'),
+  overriddenRisk: DS.attr('string'),
```
cosmosgenius: This is suppose to be number? not string its the same as risk param except it may have a null value.

app/helpers/analysis-status.js
```diff
@@ -6,6 +6,7 @@ export function analysisStatus(params) {
   if(typeof risk === "object") {
     risk = risk.value;
   }
+  risk = parseInt(risk);
```
cosmosgenius: is the parseInt required?

app/models/security/analysis.js
yashwin: Yes, correct. I am using ember power select here and 0 is considered as null for some reason which is why I had to make it a string.

app/helpers/analysis-status.js
yashwin: Yes, since the status is string and we are checking for int values

app/models/security/analysis.js
cosmosgenius: Please add a comment on why this change was done.

yashwin: done 👍

PR #674, opened 2018-08-17
Runtime config

yashwin:


app/initializers/irene.js
```diff
@@ -16,6 +16,40 @@ const initialize = function(application) {
   // Inject Store
   application.inject('component', 'store', 'service:store');
 
+  let runtimeConfig = window.runtimeGlobalConfig;
+
+  function parseUrl(url) {
+    let parser = document.createElement('a'), searchObject = {}, queries, split, i;
```
Praseetha-KR: instead of a tag can you use the webapi https://developer.mozilla.org/en-US/docs/Web/API/URL? this api is available in node also natively.

app/initializers/irene.js
```diff
@@ -16,6 +16,40 @@ const initialize = function(application) {
...
+    for(i = 0; i < queries.length; i++ ) {
+        split = queries[i].split('=');
+        searchObject[split[0]] = split[1];
+    }
+    return parser;
+  }
+
+  if(runtimeConfig) {
+    ENV.host = runtimeConfig.IRENE_API_HOST || ENV.host;
+    var devicefarmEnv = runtimeConfig.IRENE_DEVICEFARM_URL;
```
Praseetha-KR: devicefarmEnv should be ||'d also?

app/initializers/irene.js
```diff
@@ -16,6 +16,40 @@ const initialize = function(application) {
...
+      let deviceFarmWebsockifyHost = parseUrl(devicefarmEnv);
+      let deviceFarmSsl = deviceFarmWebsockifyHost.protocol === "wss:";
+      ENV.deviceFarmSsl = deviceFarmSsl;
+      ENV.deviceFarmPort = deviceFarmWebsockifyHost.port || (deviceFarmSsl ? 443:80);
+      ENV.deviceFarmHost = deviceFarmWebsockifyHost.hostname;
+    }
+    ENV.socketPath = runtimeConfig.IRENE_API_SOCKET_PATH || ENV.socketPath;
+    ENV.enableSSO = runtimeConfig.IRENE_ENABLE_SSO || ENV.enableSSO;
+    ENV.isEnterprise = runtimeConfig.ENTERPRISE || ENV.isEnterprise;
+    ENV.whitelabel = {
```
Praseetha-KR: Probably should be done via Object.assign. The whitelabel object may have other params which are not runtime derived, something like `ENV.whitelabel = Object.assign({}, ENV.whitelabel, <override object>)`.

Or use es6 object spread to do the same.

app/initializers/irene.js
yashwin: No, not required

app/initializers/irene.js
yashwin: Yeah, this works if we have the same key. But, here in our case - for example if white-label is enabled then it will be `ENV.whitelabel.enabled` and key in `runtimeConfig` will be `untimeConfig.WHITELABEL_ENABLED`

yashwin: if the key can be same, then we can do that 👍

app/initializers/irene.js
yashwin: done

app/initializers/irene.js
yashwin: @cosmosgenius 🔝

cosmosgenius: the override object would be something like 
```
{
	"enabled": runtimeConfig.WHITELABEL_ENABLED,
	.
	.
	.
}
```

PR #681, opened 2018-09-06
Handle connectivity issues

yashwin:
Depends on https://github.com/appknox/mycroft/pull/982


app/components/system-status/component.js
```diff
@@ -0,0 +1,46 @@
...
+import Ember from 'ember';
+import ENV from 'irene/config/environment';
+import {isNotFoundError} from 'ember-ajax/errors';
+
+export default Ember.Component.extend({
+
+  isStorageWorking: false,
+  isDeviceFarmWorking: false,
+
+  didInsertElement() {
```
cosmosgenius: Move this to didRender instead.. didInsertElement is usually used to access and hook into dom.

app/components/system-status/component.js
```diff
@@ -0,0 +1,46 @@
...
+    this.storageStatus();
+    this.deviceFarmStatus();
+  },
+
+  deviceFarmStatus() {
+    this.get("ajax").request(ENV.endpoints.ping)
```
cosmosgenius: Move this to concurrency. Along with the other function which do network calls

app/styles/app.sass
```diff
@@ -2618,3 +2618,67 @@ ul.email-list
...
+    left: 43%
+
+.system-status-container
+  margin-top: 100px
+  background: lighten($default, 50%)
+  padding-bottom: 10px
+  padding-top: 20px
+  border-radius: 10px
+  .system-status-box
+    ul
```
cosmosgenius: use class instead of tag. Ideally BEM.

PR #688, opened 2018-09-27
Add analytics feature request & update account ID, user ID

yashwin:


config/environment.js
```diff
@@ -265,7 +268,7 @@ module.exports = function(environment) {
     ENV['ember-cli-mirage'] = {
       enabled: false
     };
-    ENV['host'] = "http://localhost:8000";
+    ENV['host'] = "https://api.appknox.com";
```
cosmosgenius: won't this cause issue?

PR #698, opened 2018-10-04
Implement toggling api scan in security dashboard

yashwin:
Depends on - https://github.com/appknox/mycroft/pull/998

### DIFF
--- a/app/components/file-actions.js
+++ b/app/components/file-actions.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import ENUMS from 'irene/enums';
 import { task } from 'ember-concurrency';
 import { on } from '@ember/object/evented';
 import ENV from 'irene/config/environment';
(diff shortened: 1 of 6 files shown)


app/components/file-actions.js
```diff
@@ -29,6 +30,29 @@ export default Ember.Component.extend({
     return [ireneHost, "file", fileId].join('/');
   }).property(),
 
+  setApiScanStatus: task(function * () {
+    let isApiDone = this.$("#api-scan-status").is(":checked");
```
cosmosgenius: Can't we do this via binding? instead of accessing it directly here?

app/components/file-actions.js
```diff
@@ -29,6 +30,29 @@ export default Ember.Component.extend({
...
+    let apiScanStatus = ENUMS.SCAN_STATUS.UNKNOWN;
+    if(isApiDone) apiScanStatus = ENUMS.SCAN_STATUS.COMPLETED;
+    const file = yield this.get("fileDetails")
+    yield file.set('apiScanStatus', apiScanStatus);
```
cosmosgenius: also set can't be yield'd

PR #701, opened 2018-10-09
Implementation of Analytics(Stat) Page

yashwin:


app/components/overall-report.js
```diff
@@ -2,45 +2,9 @@ import Ember from 'ember';
...
-  //         "#EF4836",
-  //         "#FF8C00",
-  //         "#F5D76E",
-  //         "#2CC2F8"
-  //       ]
-  //     } ]
-  //   };
-  // }).property("stat.totalCritical", "stat.totalHigh", "stat.totalMedium", "stat.totalLow")
+  analytics: (function() {
+    return this.get('store').findRecord('analytics', 1);
```
cosmosgenius: Here better would be query record i think.

app/components/overall-report.js
```diff
@@ -2,45 +2,9 @@ import Ember from 'ember';
...
+  }).property()
```
cosmosgenius: Also let move away from `.property` to Ember.computed

app/serializers/analytics.js
```diff
@@ -0,0 +1,21 @@
+import DRFSerializer from './drf';
+
+export default DRFSerializer.extend({
+  normalizeResponse: function (store, primaryModelClass, payload) {
```
cosmosgenius: Lets not manually serialise here!

app/routes/authenticated/index.js
```diff
@@ -2,13 +2,16 @@ import Ember from 'ember';
...
+  session: service(),
+
   title: `Home${config.platform}`,
+
   model() {
-    this.modelFor("authenticated");
-  },
-  beforeModel() {
-    this.transitionTo('/projects');
+    const userId = this.get("session.data.authenticated.user_id");
```
cosmosgenius: Do we need to do this? returning `this.modelFor("authenticated");` should work?

PR #702, opened 2018-10-09
Added download excel report in security dashboard

Praseetha-KR:
##### PREVIEW:

screenshot 2018-10-09 at 7 29 18 pm

### DIFF
--- a/app/templates/components/file-actions.emblem
+++ b/app/templates/components/file-actions.emblem
@@ -17,6 +17,17 @@
             | Generate Report
           a.margin-left href=ireneFilePath target="_blank" class="button is-primary"
             | Visit Dashboard
+          span.dropdown.dropdown--default.margin-left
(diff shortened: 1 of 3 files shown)


app/templates/components/file-actions.emblem
```diff
@@ -17,6 +17,17 @@
             | Generate Report
           a.margin-left href=ireneFilePath target="_blank" class="button is-primary"
             | Visit Dashboard
+          span.dropdown.dropdown--default.margin-left
+            = basic-dropdown horizontalPosition='right' as |dropdown|
```
yashwin: I think we can use `attach-tooltip` here instead of custom drop down

something like `= attach-tooltip placement="top"`

We are doing similar thing during logout

Praseetha-KR: We would require dropdown menu here which has a clickable action, tooltip is not what we needed here.

PR #730, opened 2019-02-22
Set user company to crisp

yashwin:
Depends on https://github.com/appknox/mycroft/pull/1046

### DIFF
--- a/app/routes/authenticated.js
+++ b/app/routes/authenticated.js
@@ -42,7 +42,11 @@ const AuthenticatedRoute = Route.extend(AuthenticatedRouteMixin, {
       accountId: this.get("org.selected.id")
     };
     triggerAnalytics('login', data);
-    chat.setUser(user.get("email"), user.get("crispHash"))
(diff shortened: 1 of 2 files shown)


app/routes/authenticated.js
```diff
@@ -42,7 +42,10 @@ const AuthenticatedRoute = Route.extend(AuthenticatedRouteMixin, {
       accountId: this.get("org.selected.id")
     };
     triggerAnalytics('login', data);
-    chat.setUser(user.get("email"), user.get("crispHash"))
+    chat.setUserEmail(user.get("email"), user.get("crispHash"))
+    const org = this.get("org.selected.data.name")
+    const company = org || user.get("email").split("@").pop().trim()
```
subho007: `user` also has a company field, check that first.

subho007: `user.company`

yashwin: oh, that is not beings sent in the API response, will add that and use it

PR #731, opened 2019-02-25
Disable register page and move the link to new registration page

cosmosgenius:


config/environment.js
```diff
@@ -1,5 +1,10 @@
 var url = require('url');
 
+function isTrue(value) {
+  value = String(value).toLowerCase();
+  return value === 'true';
+}
+
```
Praseetha-KR: @cosmosgenius why not make this a helper/util?

cosmosgenius: May be will move when we used it ones more

PR #733, opened 2019-02-27
Upgrade novnc

yashwin:
This needs to be tested properly

### DIFF
--- a/app/components/vnc-viewer.js
+++ b/app/components/vnc-viewer.js
@@ -1,10 +1,10 @@
 import Component from '@ember/component';
 import { inject as service } from '@ember/service';
 import { computed, observer } from '@ember/object';
-import { isEmpty } from '@ember/utils';
 import ENUMS from 'irene/enums';
 import ENV from 'irene/config/environment';
 import { translationMacro as t } from 'ember-i18n';
+import RFB from '@novnc/novnc/core/rfb';
 
 const VncViewerComponent = Component.extend({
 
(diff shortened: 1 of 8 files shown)


app/templates/components/vnc-viewer.emblem
```diff
@@ -12,7 +12,7 @@ if isPoppedOut
         if isNotTablet
           div class="speaker"
       div class=screenRequired:noscreen:screen
-        canvas.canvas
+        div.canvas
```
subho007: name the div as `screen` or something rather than canvas. It is very confusing

app/styles/app.sass
```diff
@@ -1816,8 +1816,8 @@ button.delete-button
 .marvel-device
   .screen
     box-shadow: none
-    .canvas
-      cursor: pointer!important
+    .canvas canvas
```
subho007: check my other message regarding the name

app/components/vnc-viewer.js
```diff
@@ -30,40 +29,32 @@ const VncViewerComponent = Component.extend({
   }),
 
   setupRFB() {
-    const rfb = this.get("rfb");
-    if (!isEmpty(rfb)) {
-      return;
-    }
     const canvasEl = this.element.getElementsByClassName("canvas")[0];
```
subho007: Hook into the element named `screen` rather than canvas

app/components/vnc-viewer.js
```diff
@@ -30,40 +29,32 @@ const VncViewerComponent = Component.extend({
...
-      'onUpdateState'() {
-        setTimeout(that.set_ratio.bind(that), 500);
-        return true;
-      },
-
-      'onXvpInit'() {
-        return true;
+    const deviceToken = this.get("file.deviceToken");
+    const deviceFarmURL =
+      `wss://devicefarm2.appknox.com/websockify?token=${deviceToken}`
```
subho007: should not be hardcoded, take it from the config vars

app/components/vnc-viewer.js
```diff
@@ -30,40 +29,32 @@ const VncViewerComponent = Component.extend({
...
+
+    const rfb = new RFB(
+      canvasEl,
+      deviceFarmURL, {
+        'credentials': {
+            'password': '1234'
```
subho007: take it from env variable

app/components/vnc-viewer.js
```diff
@@ -30,40 +29,32 @@ const VncViewerComponent = Component.extend({
...
-    })
     );
-
-    if (this.get('file.isReady')) {
-      return this.send("connect");
-    }
+    rfb.addEventListener('connect', () => {
+      const platform = this.get("file.project.platform");
+      if (platform === ENUMS.PLATFORM.IOS) {
+        rfb.scaleViewport = true;
```
subho007: why not `scaleViewport = true` for both Android and iOS ?

app/templates/components/vnc-viewer.emblem
yashwin: div as screen? screen is not an element. we can rather name the class as screen, but already have screen which contains the canvas. we can name the class as canvas-container or something like that

app/components/vnc-viewer.js
yashwin: If we set `scaleViewport=true` for android, then the scales to half of the screen. This is only needed for iOS devices

subho007:
@Praseetha-KR @cosmosgenius needs your reviews on this. @yashwin is this completed?

subho007:
resolve the conflicts @yashwin

subho007:
@cosmosgenius any update on this ?

cosmosgenius:
Closing this...Fixed in another PR

PR #734, opened 2019-03-04
VNC typing issue

yashwin:
Keyboard hits are not registered while dynamic scan is running, this PR resolves the issue


subho007:
@yashwin when setting focus first clear the focus explicitly, this makes sure even if by mistake the `blurfocus` was not called previously, it blurs it and sets it. Am I right @Praseetha-KR ?

yashwin:
@subho007 Right, even I was thinking the same a while ago and saw your comment here 😆

PR #735, opened 2019-03-08
Hide chat for unauthenticated users

yashwin:


subho007:
@Praseetha-KR @cosmosgenius check if this is a good way to hide the chat button for unauthenticated users

PR #743, opened 2019-04-03
Fix: Attachment file issue

RaiVaibhav:
Previously during uploading attachment the whole analysis component get's update which which reseting the input component value to the value save in backend.

### DIFF
--- a/app/components/edit-analysis-details.js
+++ b/app/components/edit-analysis-details.js
@@ -192,7 +192,6 @@ export default Component.extend({
       availability_impact: availabilityImpact
     };
     const url = [ENV.endpoints.analyses, analysisid].join('/');
-    
     yield this.get("ajax").put(url,{ namespace: 'api/hudson-api', data: JSON.stringify(data), contentType: 'application/json' })
   }),
 
--- a/app/components/edit-analysis-details.js
+++ b/app/components/edit-analysis-details.js
@@ -5,6 +5,7 @@ import { isEmpty } from '@ember/utils';
 import { getOwner } from '@ember/application';
 import ENUMS from 'irene/enums';
 import ENV from 'irene/config/environment';
+import { task } from 'ember-concurrency';
 
 
 export default Component.extend({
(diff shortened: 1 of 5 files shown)


app/components/edit-analysis-details.js
```diff
@@ -127,18 +128,85 @@ export default Component.extend({
...
       const fileName = file.blob.name;
       const data = {
         name: fileName
       };
       const analysisid= this.get("analysis.analysisid");
       try {
-        var fileData = await this.get("ajax").post(ENV.endpoints.uploadFile,{namespace: 'api/hudson-api', data});
-        await file.uploadBinary(fileData.url, {
+        var fileData = yield this.get("ajax").post(ENV.endpoints.uploadFile,{namespace: 'api/hudson-api', data});
+        file.uploadBinary(fileData.url, {
```
cosmosgenius: this needs to be yielded

app/components/edit-analysis-details.js
```diff
@@ -149,19 +217,27 @@ export default Component.extend({
...
         };
-        await this.get("ajax").post(ENV.endpoints.uploadedAttachment,{namespace: 'api/hudson-api', data: fileDetailsData});
-
-        this.set("isUploading", false);
-        this.get("notify").success("File Uploaded Successfully");
-        const analysisObj = this.get("store").findRecord('security/analysis', this.get("analysis.analysisid"));
-        this.set('analysisDetails', analysisObj);
+        yield this.get("ajax").post(ENV.endpoints.uploadedAttachment,{namespace: 'api/hudson-api', data: fileDetailsData});
+        yield this.get('detailSaveAjaxCallUtil').perform()
+        .then(() => {
```
cosmosgenius: Don't use then here use the normal code to continue execution

app/templates/components/edit-analysis-details.emblem
```diff
@@ -191,9 +191,9 @@
         .column
           form
             p.control.is-fullwidth
-              = file-upload name="files" accept="*/*" onfileadd=(action "uploadFile") as |queue|
-                a class="button is-primary" disabled=isUploading
-                  if isUploading
+              = file-upload name="files" accept="*/*" onfileadd=( action 'uploadFileWrapper' ) as |queue0|
```
cosmosgenius: queue0?

app/components/edit-analysis-details.js
```diff
@@ -127,18 +128,85 @@ export default Component.extend({
...
   },
 
-  actions: {
 
-    async uploadFile(file) {
+
+  detailSaveUtil: task(function * (param){
+    let key=param;
+    yield this.get('detailSaveAjaxCallUtil').perform(param)
+    .then(() => {
```
cosmosgenius: Don't use then there as well use normal code. You can use normal try catch to catch errors.

app/templates/components/edit-analysis-details.emblem
RaiVaibhav: @cosmosgenius  will fix all :+1: , this one I think gets added by mistake during committing the changes, thanks for the review  :+1:

app/components/edit-analysis-details.js
```diff
@@ -127,18 +128,85 @@ export default Component.extend({
...
-  actions: {
 
-    async uploadFile(file) {
+
+  detailSaveUtil: task(function * (param){
+    let key=param;
+    try{
+      yield this.get('detailSaveAjaxCallUtil').perform(param);
+      this.get("notify").success("Analyses Updated");
+      if(key) {
```
Praseetha-KR: Better to do `key == 'back'` check

PR #744, opened 2019-04-04
Show projects menu on start and add analytics route

yashwin:


app/templates/components/home-page.emblem
```diff
@@ -29,18 +29,18 @@ if isLoaded
...
-                  = t "dashboard"
             li
               = link-to 'authenticated.projects' class="mp-auth-projects"
                 div
                   = fa-icon "tasks" class="sidebar-icons"
                   = t "projects"
                 span.badge-square.is-sidebar.content-right
                   | #{organization.selected.projectsCount}
+            li
+              = link-to 'authenticated.analytics' class="mp-auth-index"
```
Praseetha-KR: What's the use of `mp-auth-*` class? shouldn't it be `mp-auth-analytics`?

yashwin: Its for mixpanel

PR #761, opened 2019-05-03
Initial POC of multiple dynamicscan

RaiVaibhav:


guganabu:
I think, it's not valid an will be part of dynamic scan timeline!
if so, we could close it

subho007:
Closing the PR but keeping the Branch for reference 👍

PR #765, opened 2019-05-14
Show min OS version for app in dynamic scan modal

Praseetha-KR:
Screenshot 2019-05-15 at 11 45 17 PM

### DIFF
--- a/app/templates/components/dynamic-scan.emblem
+++ b/app/templates/components/dynamic-scan.emblem
@@ -53,46 +53,64 @@ if notVNCViewer
       .api-scanning
         h6
           = t "modalCard.runAPIScan.description"
-        .project-preference-component
+        .project-preference-component.margin-v-1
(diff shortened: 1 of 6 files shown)


app/templates/components/dynamic-scan.emblem
```diff
@@ -72,10 +72,15 @@ if notVNCViewer
...
         div.warning-band.info-band
           .columns
-            .column.is-one-sixth
+            .column.is-one-sixth.align-middle
               = fa-icon 'info-circle'
             .column
               = t "modalCard.apiScan.dynamicScanText"
+        if file.minOsVersion
+          .alert.alert-info.margin-t-3q
+            | App requires minimum OS version
```
subho007: @Praseetha-KR this needs to be added in locales so that we can get a Japanese translation of the message

PR #773, opened 2019-06-19
Add: Proxy Settings

RaiVaibhav:


app/templates/components/dynamic-scan.emblem
```diff
@@ -124,6 +160,33 @@ if notVNCViewer
               = fa-icon 'exclamation-triangle'
             .column
               = t "modalCard.apiScan.extraDescription"
+        .margin-top20.add-filter-list
+          .columns
+            .column.bold.black-text.margin-bottom-5
+              | Proxy Settings
```
Praseetha-KR: All strings used here should be translated

app/styles/app.sass
```diff
@@ -3312,3 +3316,19 @@ $filetag-bg: #f3f5f8
   &-item
     height: 1.6em
     border-radius: 0.2em
+
+.flex-2
+  flex: 2
+.flex-3
+  flex: 3
+.warning-msg
+  height: 30px;
```
Praseetha-KR: Remove `;`

app/styles/app.sass
```diff
@@ -3312,3 +3316,19 @@ $filetag-bg: #f3f5f8
...
+
+.margin-bottom-5
+  font-weight: 900;
+  font-size: 18px;
+  margin-bottom: 5px;
```
Praseetha-KR: For bottom margins use already existing `margin-b-*` classes. 

Also, util class with name `margin-bottom-5` should never contain other styles like font style modifiers, use specific class for the target element or use existing util classes like `font-medium`, `bold` etc.

app/models/file.js
```diff
@@ -215,7 +215,11 @@ const File = DS.Model.extend(BaseModelMixin, {
 
   setReady() {
     this.setDynamicStatus(ENUMS.DYNAMIC_STATUS.READY);
-  }
+  },
+  postProxySetting(data, profileId) {
+    var adapter = this.store.adapterFor(this.constructor.modelName);
+    return adapter.postProxySetting(this.store, this.constructor.modelName, this, data, profileId);
+  },
```
Praseetha-KR: Why is this `postProxySetting` required?

app/components/dynamic-scan.js
```diff
@@ -5,26 +5,124 @@ import ENUMS from 'irene/enums';
...
+      if(changeset.get('errors') && changeset.get('errors')[0].validation) {
+        this.get("notify").error(
+          changeset.get('errors')[0].validation[0],
+          ENV.notifications
+        );
+      }
+      return;
+    }
+    let currentProxy = yield this.get('currentProxy');
+    const enabled = !currentProxy.get('enabled');
```
Praseetha-KR: this `!` is confusing, better would be to use `toggleProxy` as function name

PR #789, opened 2019-08-01
Upgrade: noVNC

RaiVaibhav:


app/components/novnc-rfb/component.js
```diff
@@ -0,0 +1,34 @@
+import Component from '@ember/component';
+import ENV from 'irene/config/environment';
+import RFB from '@novnc/novnc/core/rfb';
+
+export default Component.extend({
+  rfb: null,
+  didInsertElement(){
+    const canvasEl = this.element.getElementsByClassName("canvas-container")[0];
+    const deviceToken = this.get("file.deviceToken");
```
cosmosgenius: don't send file...send the token alone

app/components/novnc-rfb/component.js
```diff
@@ -0,0 +1,34 @@
...
+    const deviceFarmURL =
+      `wss://${ENV.deviceFarmHost}/${ENV.deviceFarmPath}?token=${deviceToken}`
```
cosmosgenius: or send the url rather

app/components/novnc-rfb/component.js
```diff
@@ -0,0 +1,34 @@
...
+
+    const rfb = new RFB(
+      canvasEl,
+      deviceFarmURL, {
+        'credentials': {
+            'password': ENV.deviceFarmPassword
```
cosmosgenius: send password also via component param

ember-cli-build.js
```diff
@@ -64,18 +64,18 @@ module.exports = function(defaults) {
...
-   app.import('bower_components/noVNC/core/input/keysymdef.js');
-   app.import('bower_components/noVNC/core/input/xtscancodes.js');
-   app.import('bower_components/noVNC/core/input/util.js');
-   app.import('bower_components/noVNC/core/util.js');
-   app.import('bower_components/noVNC/core/input/devices.js');
-   app.import('bower_components/noVNC/core/display.js');
-   app.import('bower_components/noVNC/core/inflator.js');
-   app.import('bower_components/noVNC/core/rfb.js');
-   app.import('bower_components/noVNC/core/input/keysym.js');
+   // app.import('bower_components/noVNC/core/base64.js');
```
cosmosgenius: delete these if not required

app/components/novnc-rfb/component.js
cosmosgenius: Also do not generate the url. It can the use ws or. have a port number etc. Check if the url can be gotten from config directly and append just the token

app/initializers/irene.js
```diff
@@ -26,12 +26,9 @@ const initialize = function(application) {
...
     ENV.host = runtimeConfig.IRENE_API_HOST || ENV.host;
     var devicefarmEnv = runtimeConfig.IRENE_DEVICEFARM_URL;
+    var deviceFarmPath = "websockify";
     if(devicefarmEnv) {
-      let deviceFarmWebsockifyHost = new URL(devicefarmEnv);
-      let deviceFarmSsl = deviceFarmWebsockifyHost.protocol === "wss:";
-      ENV.deviceFarmSsl = deviceFarmSsl;
-      ENV.deviceFarmPort = deviceFarmWebsockifyHost.port || (deviceFarmSsl ? 443:80);
-      ENV.deviceFarmHost = deviceFarmWebsockifyHost.hostname;
+      ENV.deviceFarmURL = `${devicefarmEnv}/${deviceFarmPath}`
```
cosmosgenius: use new URL to validate the url

cosmosgenius: use new URL to generate the url rather than appending. The env value may already contain `/`

app/styles/app.sass
```diff
@@ -2120,7 +2118,13 @@ button.delete-button
     height: 50px
     left: 52%
 
+.novnc-rfb
```
cosmosgenius: move this to local class

config/environment.js
```diff
@@ -95,10 +92,8 @@ module.exports = function(environment) {
       allowEmpty: true, // default: false
       includeLocales: ['en', 'ja']
     },
-    deviceFarmSsl: deviceFarmSsl,
-    deviceFarmPort: deviceFarmPort,
-    deviceFarmPath: "websockify",
-    deviceFarmHost: deviceFarmHost,
+    deviceFarmURL: `${devicefarmEnv}/${deviceFarmPath}`,
```
cosmosgenius: same here

app/styles/app.sass
RaiVaibhav: `novnc-rfb` is used both at the component (novnc-rfb) level class and the child div inside the parent component (novnc-rfb) and both needs width and height `100%`, so I move this to local class then I am not getting how to bind the local class with the component in ember

RaiVaibhav: Netlify is failing with the log, `URL not defined`, I am building the app locally and it's building without any errors.

config/environment.js
```diff
@@ -7,10 +7,18 @@ function isTrue(value) {
...
+  var deviceFarmPath = "/websockify";
   var deviceFarmWebsockifyHost = url.parse(devicefarmEnv);
   var deviceFarmSsl = deviceFarmWebsockifyHost.protocol == "wss:";
   var deviceFarmPort = deviceFarmWebsockifyHost.port || (deviceFarmSsl ? 443:80);
   var deviceFarmHost = deviceFarmWebsockifyHost.hostname;
+  var deviceFarmURL = url.format(
+    {
+      protocol: deviceFarmWebsockifyHost.protocol,
+      hostname: deviceFarmHost,
+      port: (deviceFarmSsl ? 443:80),
```
Praseetha-KR: If the port is specified along with host, this will not accept that. Use `deviceFarmPort` here

PR #791, opened 2019-08-08
Fix: Jira and GitHub error message

RaiVaibhav:


app/templates/components/jira-project.emblem
```diff
@@ -44,6 +44,13 @@ else
             button.is-primary click={(perform editJiraRepoModal)}
               = t "selectProject"
       else
+        if reconnect
+          h6
+            | Jira
```
cosmosgenius: no hardcoded string

cosmosgenius: no hardcoded string

PR #804, opened 2019-09-05
Initialize pendo when user account has a project

RaiVaibhav:


subho007:
Pendo initialization should be based on a new ENV

Pendo State should be = New Env if New Env is set else set to not of Enterprise Env else false

@cosmosgenius

config/environment.js
```diff
@@ -355,5 +355,14 @@ module.exports = function(environment) {
...
   }
+
+  if (process.env.IRENE_ENABLE_PENDO){
+    ENV.enablePendo = process.env.IRENE_ENABLE_PENDO;
+  }else if(environment==='whitelabel'){
+    ENV.enablePendo = !ENV.enablePendo;
+  }else{
+    ENV.enablePendo = false;
+  }
+
```
RaiVaibhav: @cosmosgenius  I don't know why you actually wanted this wired `if-else` nesting,
It can be, so what will happen is if there is no `IRENE_ENABLE_PENDO` then environment other than `whitelabel` it will set the enablePendo to `false`.

RaiVaibhav: Either it should be 
```
  if (process.env.IRENE_ENABLE_PENDO){
    ENV.enablePendo = process.env.IRENE_ENABLE_PENDO;
  }else{
    ENV.enablePendo = (environment!=='whitelabel')
  }
``

cosmosgenius: enablePendo should be set to IRENE_ENABLE_PENDO if and only if IRENE_ENABLE_PENDO is set. else it should be set to NOT of Enterprise env if and only if Enterprise Env is set. else to false.

Also can you change IRENE_ENABLE_PENDO to IRENE_PENDO_ENABLE.

cosmosgenius:
fixed in #824

PR #809, opened 2019-09-20
Enable htmlSafe to renders an ElementNode

RaiVaibhav:


Praseetha-KR:
@RaiVaibhav This needs to be added for  `uploadNewProject`, `capturedApiEmptyDesc`, `manual.enterLoginDetails` also

PR #825, opened 2019-12-26
Implement a user interface for organization archive generation and download feature.

Parthipan-Natkunam:
#### JIRA Story Id: 

- [AP-442](https://appknox.atlassian.net/browse/AP-442)

#### JIRA Task Ids:
- [AP-446](https://appknox.atlassian.net/browse/AP-446)
- [AP-447](https://appknox.atlassian.net/browse/AP-447)

#### Change Log:
1. Add ember components required for the presentation and generation of archives.
2. Create an archive model.
3. Set `usePods` to `true` in ember CLI config file.


app/styles/app.scss
```diff
@@ -0,0 +1,2 @@
+
+@import 'ember-basic-dropdown';
```
Praseetha-KR: This file is not required right?

app/models/organization-archive.js
```diff
@@ -0,0 +1,69 @@
...
+      try{
+        const response = await adapter.getDownloadURL(this.get('id'));
+        if(response && response.url && response.url.length > 0){
+          URL =  response.url;
+        }
+        return URL;
+      }catch(err){
+        debug('Download organization archive URL network call failed');
+        return URL;
+      }
```
Praseetha-KR: Can the var `URL` be removed? (also, empty string instead of `null`)
```javascript
try {
   ...
   if (response && response.url && response.url.length) {
      return response.url;
   }
} catch(err) {
   ...
}
return '';
```

app/models/organization-archive.js
```diff
@@ -0,0 +1,69 @@
...
+      const adapter = this.store.adapterFor(this.constructor.modelName);
+      let URL = null;
+      try{
+        const response = await adapter.getDownloadURL(this.get('id'));
+        if(response && response.url && response.url.length > 0){
+          URL =  response.url;
+        }
+        return URL;
+      }catch(err){
+        debug('Download organization archive URL network call failed');
```
Praseetha-KR: Send this error to rollbar instead of debug:
```javascript
logger: service('rollbar'),
...
   } catch(err) {
		this.get('logger').error('Download organization archive URL network call failed', err);
   }
...
```

PR #834, opened 2020-04-27
add API host status check feature in /status route

Parthipan-Natkunam:
**Type:** Feature

**Description:**
Adding a status check feature for the backend API Host in the '/status' route, in addition to the existing Storage and Devicefarm status checks.

**Change Log:**
1. Add status check feature for API host.
2. Improve the UX by adding a checking indicator by utilising generator's `isRunning` method
3. Remove redundant CSS rules and replace with a concise rule.
4. Auto format code with prettier in all the files which have been modified for this feature. 

**Preview:**
Screenshot 2020-04-27 at 7 43 48 PM


subho007:
I like this @Parthipan-Natkunam 👍

app/styles/app.sass
```diff
@@ -2905,7 +2905,7 @@ input:checked + .slider-nolabel:before
 
 $login-button-width: 100%
 .auth-main-container
-  height: 100vh
+  min-height: 100vh
```
Praseetha-KR, line 2908: Any reason to change this? This would break vertical centering of the login form

app/components/system-status/component.js
```diff
@@ -1,38 +1,48 @@
...
       this.set("isDeviceFarmWorking", true);
-    } catch(error) {
+    } catch (error) {
       this.set("isDeviceFarmWorking", false);
     }
   }).drop(),
 
+  getAPIHostStatus: task(function* () {
+    try {
+      yield this.get("ajax").request(`${ENV.host}/api/check`);
```
Praseetha-KR, line 42: Can you use `/api/ping`? And set it in `ENV.endpoints` rather than hardcoding here

app/components/system-status/styles.scss
```diff
@@ -1,38 +1,35 @@
...
-
-.devicefarm-operational .devicefarm-status {
-    color: $green;
-    .operational-text {
-      display: inline;
-    }
-    .unavailable-text {
-      display: none;
-    }
+.operational-status.online {
```
Praseetha-KR, line 9: Nice! 👍

app/components/system-status/component.js
```diff
@@ -1,38 +1,48 @@
...
   isStorageWorking: false,
   isDeviceFarmWorking: false,
+  isAPIHostWorking: false,
 
   localClassNameBindings: [
-    'isStorageWorking:storage-operational',
-    'isDeviceFarmWorking:devicefarm-operational'],
+    "isStorageWorking:storage-operational",
+    "isDeviceFarmWorking:devicefarm-operational",
+  ],
```
Praseetha-KR, line 14: This is unused, right?

app/styles/app.sass
Parthipan-Natkunam: The status container was longer with the new data and was breaking the layout in /scans. Hence made it to minHeight. Have checked the login form and since it doesn't have any content below the form, the vertical centering doesn't seem to be affected. Will there be any other side effects apart from that?

Parthipan-Natkunam: My bad. It is affecting the centering.. will fix it.

Parthipan-Natkunam: Have reverted this change as it is no longer required.

app/components/system-status/component.js
Parthipan-Natkunam: Done

app/components/system-status/component.js
Parthipan-Natkunam: Yes. Have removed this in the latest commit.

PR #835, opened 2020-04-28
add API host status check feature in /status route

Parthipan-Natkunam:
Due to merge issues the earlier merged PR  (#834) with this content was reverted.
So making a new PR for the same.
Since the root cause for the issue is not known yet, a branch deployment on Netlify and testing on the above said deployment is required before merging this PR.


Praseetha-KR:
_(Due to brach name conflict, I was unable to do branch deploy for #834. That's why temporarily deleted this branch. The issue is not related to code of #834 & #835, root cause is incorrect merging)_

Here is the temp [preview](https://irene-api-host-status-check.irenestaging.appknox.io/status), I think its good to go 👍(there is a small spacing at the bottom though 😄). @Parthipan-Natkunam can you verify once more? If everything is fine, go ahead & deploy 👍

Parthipan-Natkunam:
Thank you @Praseetha-KR . I have verified it on the preview. So going ahead with merging and deployment.

PR #846, opened 2020-08-07
remove phone support from registration and migrate invite-register to glimmer

cosmosgenius:


app/components/invite-register/component.js
```diff
@@ -1,87 +1,88 @@
...
+    this.registerTask.perform(changeset);
+  }
 
-});
+  @action async geoIpLookupFunc(callback) {
+    const country = await this.locationinfo.getCountry();
+    callback(country);
+    return country;
+  }
+}
```
Praseetha-KR, line 88: @cosmosgenius existence of this code is questionable 😂

PR #852, opened 2020-10-01
AP-631: Added x at end of text input

guganabu:


guganabu:
merging, since the impact is not major

PR #853, opened 2020-10-01
Three bounce spinner version

guganabu:


app/components/three-bounce-spinner/styles.scss
```diff
@@ -0,0 +1,42 @@
...
+    animation: sk-bouncedelay 1.4s infinite ease-in-out both;
+  }
+
+  .bounce1 {
+    -webkit-animation-delay: -0.32s;
+    animation-delay: -0.32s;
+  }
+
+  .bounce2 {
+    -webkit-animation-delay: -0.16s;
```
cosmosgenius, line 24: it the webkit specific css required? I think its generated via plugin. Can you remove it and test once?

cosmosgenius (changes requested):

app/components/three-bounce-spinner/styles.scss
guganabu: It may work without -webkit for recent browsers but, I just thought have it for older version support.

cosmosgenius (approved):

PR #855, opened 2020-10-09
Add ENVHandler class to handle env and runtime injection

cosmosgenius:
- Adds a hard check on ENV access by irene.
- Adds a unified way to retrieve ENV from both `process.env` and `runtimeConfig`
- Adds a function to get plugin values from ENV with dependency on `Enterprise`

### DIFF
--- a/app/initializers/irene.js
+++ b/app/initializers/irene.js
@@ -1,81 +1,152 @@
-import ENV from 'irene/config/environment';
+import ENV from "irene/config/environment";
 
-const runtimeConfig = window.runtimeGlobalConfig;
-
-function isTrue(value) {
-  value = String(value).toLowerCase();
-  return value === 'true';
-}
-
-
-function getPluginActivationStatus(pluginName){
-  const pluginEnvVariable = ENV.thirdPartyPluginEnvMap[pluginName];
-
-  if(pluginEnvVariable.env in runtimeConfig){
-    return isTrue(runtimeConfig[pluginEnvVariable.env]);
+class ENVHandler {
+  constructor(envHandlerConst) {
+    this.envHandlerConst = envHandlerConst;
   }
-  if('ENTERPRISE' in runtimeConfig){
-    return !isTrue(runtimeConfig.ENTERPRISE);
+
+  isRuntimeAvailable() {
+    return !(typeof runtimeGlobalConfig == "undefined");
   }
-  return pluginEnvVariable.default;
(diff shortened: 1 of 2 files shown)


app/initializers/irene.js
```diff
@@ -1,25 +1,91 @@
...
+  constructor(envHandlerConst) {
+    this.envHandlerConst = envHandlerConst;
+  }
 
-function isTrue(value) {
-  value = String(value).toLowerCase();
-  return value === 'true';
-}
+  isRuntimeAvailable() {
+    return !(typeof runtimeGlobalConfig == "undefined");
```
guganabu, line 9: how does this variable becomes accessible?

config/environment.js
```diff
@@ -1,71 +1,201 @@
...
+  ],
+  processENV: Object.keys(process.env).reduce((acc, key) => {
+    if (possibleENVS.indexOf(key) > -1) {
+      acc[key] = process.env[key];
+    }
+    return acc;
+  }, {}),
+};
+
+class ENVHandler {
```
guganabu, line 62: both initializer and env has the same ENVHandler class implementation, is that required to be separated?

config/environment.js
```diff
@@ -103,267 +233,429 @@ module.exports = function(environment) {
...
-      routeIfAlreadyAuthenticated: 'authenticated.index',
+    "ember-simple-auth": {
+      authenticationRoute: "/login",
+      loginEndPoint: "/login",
+      checkEndPoint: "/check",
+      logoutEndPoint: "/logout",
+      routeAfterAuthentication: "authenticated.index",
+      routeIfAlreadyAuthenticated: "authenticated.index",
     },
     endpoints: {
```
guganabu, line 309: is this really active endpoints?

guganabu (commented):
Please take a look into the comments which I have added

app/initializers/irene.js
cosmosgenius: runtimeGlobalConfig is injected via a script as global variable. See runtimeconfig.js

config/environment.js
cosmosgenius: nope, i have no idea how to unify them. I need ENVHandler in both environment.js and in initializer, but didn't find a way to share it from a single place.

guganabu (approved):

Praseetha-KR (approved):

PR #859, opened 2020-10-30
AP-646: File report button state handler

guganabu:


app/components/file/report-btn/component.js
```diff
@@ -0,0 +1,278 @@
+import Component from '@glimmer/component';
+import {
+  tracked
+} from '@glimmer/tracking';
+import {
+  task
+} from 'ember-concurrency';
```
Praseetha-KR, line 7: I think prettier was not run on this file.

app/components/file-header/template.emblem
```diff
@@ -120,22 +120,12 @@
...
-                    | &nbsp;
-                else
-                  = fa-icon "download"
-                | &nbsp; #{ t "pdfReport"}
+      if organization.selected.isTrial
+        button.is-primary.margin-top-5 click={(perform openRequestAccessModal)}
+          = fa-icon "download"
+          | &nbsp; #{ t "pdfReport"}
+      else
+        = file/report-btn file=file showCopyPasswordModal=(action "onShowPasswordModal")
```
Praseetha-KR, line 128: In other places it is `Parent::Child` format whereas here it is `parent/child` format. I would suggest to follow the same pattern everywhere for consistency.

app/components/file/report-btn/component.js
```diff
@@ -0,0 +1,278 @@
...
+      const data = yield adapter.downloadExternalReport(this.modelName, this.fileId)
+      if (data && data[type.format]) {
+        window.location = data[type.format];
+      } else {
+        this.notify.error(this.intl.t('noReportExists', { format: type.label }));
+      }
+    } catch (error) {
+      this.notify.error(parseError(error, this.intl.t('pleaseTryAgain')))
+    }
+  }).restartable() downloadExternalReport;
```
Praseetha-KR, line 276: It should not be called as external report - just use report.

app/components/file/report-btn/styles.scss
```diff
@@ -0,0 +1,93 @@
...
+      left: 0;
+      opacity: 0.2;
+    }
+  }
+
+  .report-list-dropdown {
+    .dd-trigger {
+      height: inherit;
+      border-radius: 0.4em;
+      margin-bottom: 5px;
```
Praseetha-KR, line 25: Again, please don't use `px` units!!

app/components/file/report-btn/template.hbs
```diff
@@ -0,0 +1,137 @@
...
+                  data-test-report='external-report-{{seq}}'>
+                  {{fa-icon type.icon}}
+                  {{t 'download'}} {{type.label}}
+                </a>
+              {{/each}}
+            </div>
+          </dd.content>
+        </BasicDropdown>
+      </div>
+    </div>
```
Praseetha-KR, line 28: Don't mix the security dashboard logic into main dashboard components. That's a bug NO!

translations/en.json
```diff
@@ -1,878 +1,891 @@
...
-  "welcome": "Welcome ",
-  "worsened": "Worsened",
-  "xmlMetadata": "XML Metadata",
-  "year": "Year",
-  "yearly": "Yearly",
-  "yes": "Yes",
-  "youCanCreateTeam": "you can create a new team by",
-  "youWillBeChargedOn": "You will be charged on",
-  "yourPlan": "Your Plan"
+	"0sharableCredits": "0 sharable credits",
```
Praseetha-KR, line 2: Why tabs instead of spaces?

app/adapters/file-report.js
```diff
@@ -0,0 +1,56 @@
...
+    const url = `${this.reportsBaseUrl}/${reportId}/${type}`;
+    const response = await fetch(url, {
+      headers: this.headers,
+      method: 'GET'
+    });
+    return response.json();
+  }
+
+  async reGenerateReport(modelName, fileId, data = {}) {
+    const url = `${this.get('host')}/${this.hudsonNamespace}/${ENV.endpoints.reports}/${fileId}`;
```
Praseetha-KR, line 35: hudson-api APIs should all be under `security`

Praseetha-KR (changes requested):

app/components/ak-modal/template.hbs
```diff
@@ -5,12 +5,12 @@
   ...attributes
 >
   {{#if @showHeader}}
-    <div local-class='modal-header'>
+    <div local-class='modal-header' data-test-modal='header'>
```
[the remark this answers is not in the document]
guganabu: ember test selectors will remove all the attr which has `data-test-*`, right? 
In earlier PR it was like data-test='' which will not be excluded from the prod build but this one should do that. Do we have any other reason than this to avoid using `data-test-*` pattern? @Praseetha-KR

app/components/file/report-btn/component.js
guganabu: Yeah, I upgraded the vscode latest version so the config seems reset. I'll fix this in next push

app/components/file-header/template.emblem
guganabu: Parent::Child is valid in octane templating, right? and I think, .emblem won't support this pattern.
please clarify...

app/components/file/report-btn/component.js
guganabu: Somewhere it was referred as external reports, which is why I used the name here. I'll change it 👍

app/components/file/report-btn/styles.scss
guganabu: okay

app/components/file/report-btn/template.hbs
guganabu: oh okie, it wasn't exposed anywhere since the beginning of the task so, it's continue to be there. Will move it to separate component then 👍

translations/en.json
guganabu: Not sure, how it's added there, will be resolved in next push 👍

app/adapters/file-report.js
guganabu: got it 👍

guganabu: okay

app/components/ak-modal/template.hbs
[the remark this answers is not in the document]
Praseetha-KR: In the [ember-test-selectors usage guide](https://github.com/simplabs/ember-test-selectors#usage), the examples provided for `data-test-*` attributes are flat unless there is a unique id. I feel it is better to keep it consistent in that way across irene rather than mixing with our own format.

Do you see any advantage of using `data-test-modal='header'` over `data-test-modal-header`?

app/components/file-header/template.emblem
Praseetha-KR: Didn't notice it is emblem, my bad 😅

app/components/ak-modal/template.hbs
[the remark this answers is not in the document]
guganabu: I don't see very drastic advantage over this. But, this could be helpful when we query group of elements like data-test-item='1', data-test-item='2', etc., yeah we can use * for pattern matching but it has more specified.
Also, I saw couple of other packages, most of them using attr with value so, I just chosen this approach.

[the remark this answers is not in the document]
Praseetha-KR: Let's stick to the documented way. Otherwise we'll have to explain to a new contributor why it was done different than the recommended way.

app/components/file/report-btn/styles.scss
Praseetha-KR: There is still lot of `px` units specified in that file even after your changes. Please fix those.

tests/integration/components/file/report-btn-test.js
```diff
@@ -0,0 +1,227 @@
...
+    assert
+      .dom(`[data-test-report="action-btn-label"]`)
+      .hasText(`t:generateReport:()`);
+
+    await click(`[data-test-report="action-btn-label"]`);
+    this.realtimeService.incrementProperty('ReportCounter');
+    assert
+      .dom(`[data-test-report="action-btn-label"]`)
+      .hasText(`t:generatingReport:()`);
+    assert.dom(`i.fa-spinner`).exists();
```
Praseetha-KR, line 90: Use data test selector for this instead of direct tag/class access. (That will help even if we change the icon pack in future.)

app/serializers/file-report.js
```diff
@@ -0,0 +1,13 @@
...
+export default DRFSerializer.extend({
+  primaryKey: 'id',
+
+  //Serialize request payload and can be extensible based on the need
+  serialize(snapshot) {
+    const payload = {
+      language: snapshot.attr('language')
+    }
+    return payload;
+  }
```
Praseetha-KR, line 12: Can you help me to understand what's the use of this?

Praseetha-KR (changes requested):

app/serializers/file-report.js
```diff
@@ -0,0 +1,5 @@
+import DRFSerializer from './drf';
+
+export default DRFSerializer.extend({
+  primaryKey: 'id',
```
Praseetha-KR, line 4: Is specifying `id` required?

Praseetha-KR (changes requested):

Praseetha-KR:
Also do a rebase

app/serializers/file-report.js
guganabu: when I initially started this task, I referred a file which has same thing so, just copied 😄

Praseetha-KR (approved):

PR #860, opened 2020-11-04
Enable crisp via website id

guganabu:


cosmosgenius:
I think you can remove `enableCrisp` and directly use and empty string check

cosmosgenius:
also env needs to start with `IRENE_` so rename `CRISP_WEBSITE_ID` to `IRENE_CRISP_WEBSITE_ID`

cosmosgenius (changes requested):
Check the initializer also

cosmosgenius (approved):

PR #876, opened 2020-12-16
Upgrade modal rendering mechanism

guganabu:


guganabu:
@cosmosgenius @Praseetha-KR I think, this is a cleanup thing, which we can take. Can you plz check it out and merge, if everything is fine?

PR #889, opened 2021-04-06
AP-688: Check file details tag char limit

guganabu:


guganabu:
merging, since impact is very low

PR #894, opened 2021-04-14
AP-694: Fixed default scroll in project card

guganabu:


subho007 (commented):
Is this the fix? I thought some other things are also bringing in the scrollbars

PR #930, opened 2021-06-11
Base theme using CSS variables & migrated partner dashboard component colors

Praseetha-KR:


guganabu:
@Praseetha-KR why shouldn't we use _vars.scss file for to define style variables?

Praseetha-KR:
@guganabu what all variables are you planning to keep in `_vars.scss` file?

> @Praseetha-KR why shouldn't we use _vars.scss file for to define style variables?

guganabu:
> @guganabu what all variables are you planning to keep in `_vars.scss` file?
> 
> > @Praseetha-KR why shouldn't we use _vars.scss file for to define style variables?

I am thinking we could use the existing _vars.scss file for to define variables related to color codes or else, may I know, these theme vars are related to whitelabel things?

Praseetha-KR:
@guganabu Ideally _vars file should have variables defined by styleuide or design system. At this point we don't have such a specification in terms of standardized margin, padding, font size, etc. rules. But we are somewhat clear about theme because of the color customization required for whitelabel themes. 

Therefore, as a starting point, I'm keeping theme related variables & classes in _theme.scss. I'm not touching the existing bohemia, app or _vars files, that cleanup is heavy and needs to be done in dedicated time. Intent of this PR is to prevent from defining random colors in components. When we have a design specification in place, it will be easier to handle theme, we can decide whether to keep color vars in _theme or in general _vars file at that point.

guganabu:
> @guganabu Ideally _vars file should have variables defined by styleguide or design system. At this point we don't have such a specification in terms of standardized margin, padding, font size, etc. rules. But we are somewhat clear about theme because of the color customization required for whitelabel themes.
> 
> Therefore, as a starting point, I'm keeping theme related variables & classes in _theme.scss. I'm not touching the existing bohemia, app or _vars files, that cleanup is heavy and needs to be done in dedicated time. Intent of this PR is to prevent from defining random colors in components. When we have a design specification in place, it will be easier to handle theme, we can decide whether to keep color vars in _theme or in general _vars file at that point.

@Praseetha-KR Cool, got it 👍

PR #931, opened 2021-06-14
Updated test attr to prefixed by data-test-

guganabu:
Updated test attributes to have the prefix of `data-test-` which will be removed from the production build by as suggested by https://github.com/simplabs/ember-test-selectors

### DIFF
--- a/app/components/partner/client-plan/template.hbs
+++ b/app/components/partner/client-plan/template.hbs
@@ -3,32 +3,32 @@
     {{#if this.getClientPlan.isRunning}}
       {{fa-icon 'circle-o-notch fa-spin text-lightgray'}}
     {{else if this.clientPlan}}
-      <div local-class='plan' data-test-payment-plan data-test='client-plan'>
+      <div local-class='plan' data-test-client-plan>
         <div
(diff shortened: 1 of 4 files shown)


app/components/partner/client-plan/template.hbs
```diff
@@ -3,32 +3,32 @@
     {{#if this.getClientPlan.isRunning}}
       {{fa-icon 'circle-o-notch fa-spin text-lightgray'}}
     {{else if this.clientPlan}}
-      <div local-class='plan' data-test-payment-plan data-test='client-plan'>
+      <div local-class='plan' data-test-payment-plan data-test-client-plan>
```
Praseetha-KR, line 6: `data-test-payment-plan` can be removed right?

Praseetha-KR (changes requested):

app/components/partner/client-plan/template.hbs
guganabu: yes, I'll remove this and check is there any dependency check for this attr

Praseetha-KR (approved):

PR #932, opened 2021-06-16
Registration integration with configuration api

cosmosgenius:
Removes two envs
- `IRENE_ENABLE_REGISTRATION`
- `IRENE_REGISTRATION_LINK`


app/components/login-component/index.scss
```diff
@@ -0,0 +1,118 @@
...
+
+.login-title {
+  font-size: 1.72rem;
+  text-align: center;
+  margin-top: 2em;
+}
+
+.login-footer {
+  margin-top: 3em;
+  border-top: 1px solid #efefef;
```
guganabu, line 31: shall we use style var here?

app/components/login-component/index.scss
```diff
@@ -0,0 +1,118 @@
...
+
+.input-label {
+  border-width: 1px 0 1px 1px;
+  border-style: solid;
+  border-color: #e6e6e6;
+  box-shadow: inset 0 1px 2px rgba(107, 107, 107, 0.1);
+  padding: 0 0.6em;
+  display: flex;
+  align-items: center;
+  background: #f7f7f7;
```
guganabu, line 95: to be consistence, all the new components are started with using style var as much as possible which is available in the theme.scss. Also, it's suggested to use em/rem values instead of px.

app/components/login-component/login/index.hbs
```diff
@@ -0,0 +1,25 @@
+<form ...attributes onsubmit="return false">
+  <div class="{{local-class 'input-group' from='irene/components/login-component/index.scss'}}">
```
guganabu, line 2: I think, we could have the file path in a prop so that it could be easy to manage, may be in future.

guganabu (commented):
Added few comments, plz share your thoughts on this!

Praseetha-KR (approved):

app/components/login-component/login/index.hbs
cosmosgenius: it will be a way more harder to manage if it was abstracted out, since the styles are not a single unit of computation rather are individual on its own.

app/components/login-component/index.scss
cosmosgenius: i don't think we can use px for `border-width`, `box-shadow` or `border-radius` since it should not scale with font-size.

app/components/login-component/index.scss
cosmosgenius: ya wanted to but...it is not defined yet...these were the original colors from previous components...once it is defined will migrate.

PR #938, opened 2021-06-21
Add integration service and migrate crisp to it

cosmosgenius:


app/templates/activate.hbs
```diff
@@ -1,12 +0,0 @@
-{{page-title "Activate"}}
-<div class="auth-main-container">
-  <AuthAssets></AuthAssets>
-  <div class="activate-container">
-    <div class="activate-main">
```
Praseetha-KR, line 5: can you remove `activate-container` & `activate-main` style definitions if not used

app/utils/install-crisp.js
```diff
@@ -1,21 +0,0 @@
-// jshint ignore: start
-import ENV from 'irene/config/environment';
-
-const installCrisp = function () {
-  if (!ENV.crispWebsiteId || ENV.crispWebsiteId === '') {
-    // eslint-disable-next-line no-console
-    return console.log("Crisp Disabled");
-  }
-  window.$crisp = [];
-  window.$crisp.push(['do', 'chat:hide']);
```
Praseetha-KR, line 10: Can you verify locally if chat container is hidden by default?

Praseetha-KR (changes requested):

app/templates/activate.hbs
cosmosgenius: removed

app/utils/install-crisp.js
cosmosgenius: default is hidden

Praseetha-KR (approved):

PR #960, opened 2021-08-27
Button UI component

guganabu:


app/components/ak-button/index.scss
```diff
@@ -0,0 +1,29 @@
+.btn {
+  height: 2.5rem;
+  padding: 0.57em 0.85em;
```
cosmosgenius, line 3: this padding number is so weird. where is this coming from?
also height should be em not rem as it should depend on the font size of the text used not root font size

app/components/ak-button/index.scss
```diff
@@ -0,0 +1,29 @@
...
+  font-size: 1rem;
+  display: flex;
+  align-items: center;
+  flex-direction: row;
+  justify-content: center;
+  column-gap: 0.5em;
+  line-height: 2.5rem;
+  cursor: pointer;
+
+  span {
```
cosmosgenius, line 15: use class there no direct reference to tag

app/components/ak-button/index.scss
```diff
@@ -0,0 +1,29 @@
...
+  padding: 0.57em 0.85em;
+  border: none;
+  border-radius: var(--button-border-radius);
+  font-size: 1rem;
+  display: flex;
+  align-items: center;
+  flex-direction: row;
+  justify-content: center;
+  column-gap: 0.5em;
+  line-height: 2.5rem;
```
cosmosgenius, line 12: are these to even required?

app/components/ak-button/index.scss
```diff
@@ -0,0 +1,29 @@
...
+  cursor: pointer;
+
+  span {
+    //Icon size
+    font-size: 1rem;
```
cosmosgenius, line 17: isn't this inherited from parent?

app/components/ak-button/neutral-outlined/index.hbs
```diff
@@ -0,0 +1,13 @@
+<button
+  type="button"
+  disabled={{@isDisabled}}
+  local-class="neutral-outlined"
+  ...attributes
+  data-test-btn-neutral-outlined
+  {{on 'click' @onClick}}
+>
+  {{#if @icon}}
```
cosmosgenius, line 9: this should not exist here...it should be yielded along with text

cosmosgenius (changes requested):

app/components/ak-button/index.scss
guganabu: 1. was tried to derive it, but having 0.5em is better 👍 
2. updated height unit

app/components/ak-button/index.scss
guganabu: yes, earlier it was a problem, now it is not anymore required!

app/components/ak-button/index.scss
guganabu: yes, it required to be overwrite with global style which is 24px set to button element. which is ideally required to align the inner content at center by pos.

app/components/ak-button/index.scss
guganabu: yep, wanted to not change default properties, but makes sense to update... and done!

app/components/ak-button/neutral-outlined/index.hbs
guganabu: It was in thought process but later on decided to wrap the contents within component itself. Moved now

cosmosgenius (approved):

PR #1010, opened 2022-04-18
fix page unresponsiveness while selecting '1 Month' in the Analytics

Yibaebi:


app/components/partner/client-uploads-stat-chart/index.hbs
```diff
@@ -48,13 +48,16 @@
               class='button'
               type='button'
               local-class='option {{if
-                (eq this.currentTimeline.key option.key)
```
cosmosgenius, line 51: This should still be key, as the point of key is to do object matching.

app/components/partner/client-uploads-stat-chart/index.js
```diff
@@ -65,10 +69,12 @@ export default class PartnerClientUploadsStatChartComponent extends Component {
       },
     },
     {
-      key: this.intl.t('monthWithNum', {
+      key: 'month',
+      axisText: this.intl.t('month'),
+      buttonSelectorText: this.intl.t('monthWithNum', {
```
cosmosgenius, line 74: can you change this to the new month inlt.

cosmosgenius (changes requested):

app/components/partner/client-uploads-stat-chart/index.js
Yibaebi: Are you referring to the axisText?

app/components/partner/client-uploads-stat-chart/index.hbs
Yibaebi: Done

app/components/partner/client-uploads-stat-chart/index.js
Yibaebi: Done

cosmosgenius (approved):

PR #1049, opened 2022-09-01
power select upgrade

Yibaebi:


app/components/partner/export-stats/index.js
```diff
@@ -55,7 +54,7 @@ export default class PartnerExportStatsComponent extends Component {
       element.click();
       document.body.removeChild(element);
     } catch (err) {
-      this.notify.error(parseError(err));
+      return;
```
cosmosgenius, line 57: Since this try is not doing anything lets remove and let the framework handle it if it fails.

cosmosgenius (changes requested):

app/components/partner/export-stats/index.js
Yibaebi: Alright. I'll tend to this.

cosmosgenius (approved):

PR #1052, opened 2022-09-06
flex changes part 1

future-pirate-king:


app/components/organization-member-list/styles.scss
```diff
@@ -2,4 +2,8 @@
   display: flex;
   align-items: center;
   justify-content: flex-end;
-}
\ No newline at end of file
+}
+
+.search-input {
+  width: 300px;
```
cosmosgenius, line 8: Use em

cosmosgenius: this has to scale with font size

cosmosgenius (changes requested):

cosmosgenius (approved):

PR #1060, opened 2022-09-22
Make the risk count a compute property

cosmosgenius:


Yibaebi (commented):
Looks fine to me. The change works as intended.

PR #1159, opened 2023-01-20
upgrade concurrency and remove decorated package

cosmosgenius:
Previous the decorated way of using concurrency was not working with non-decorator way. Now its working thus remove the decorated packages..
Tested the flow using Team create workflow where decorated is used.


cosmosgenius:
test failing with `@computed` issue closing it

PR #1193, opened 2023-02-15
TS migration of organization & me service

future-pirate-king:


app/services/me.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  async getMembership() {
+    const org = await this.queryOrganizationMe;
+    const user_id = org.id;
+
+    return this.store.findRecord('organization-member', user_id);
+  }
+
+  // get partner() {
```
Yibaebi, line 38: Commented code?

Yibaebi (approved):

app/services/me.ts
future-pirate-king: not sure the purpose of this getter as `organization-me` model don't have `partner` property, so just commented for future reference

Praseetha-KR: We can delete this @future-pirate-king , I don't think `partner` is used anywhere

future-pirate-king: yeah will do that

Praseetha-KR (approved):

PR #1204, opened 2023-02-27
org-side-menu refactor and migration

future-pirate-king:


app/components/submission-list/index.js
```diff
@@ -5,28 +5,30 @@ import { observer } from '@ember/object';
 import ENUMS from 'irene/enums';
```
Yibaebi, line 5: Is there a reason why this component class isn't migrated?

future-pirate-king: I migrated this but unit tests were not passing so reverted for now

Yibaebi (approved):

PR #1218, opened 2023-04-19
PD-813: File report updates

Yibaebi:


mirage/factories/file.js
```diff
@@ -19,6 +19,7 @@ export default Base.extend({
   isDynamicDone: faker.random.boolean(),
   isManualDone: faker.random.boolean(),
   isApiDone: faker.random.boolean(),
+  canGenerateReport: faker.random.boolean(),
```
future-pirate-king, line 22: factory keys should be have underscores instead of camel case

app/components/file/report-drawer/va-reports/index.hbs
```diff
@@ -0,0 +1,60 @@
+{{!@glint-nocheck: perform, did-insert, will-destroy}}
+<div
+  {{did-insert this.initializeDrawer}}
+  {{will-destroy this.removeReportCounterObserver}}
```
future-pirate-king, line 4: can we do this `constructor` and `willDestroy` methods

app/components/file/report-drawer/va-reports/index.ts
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  // Actions
+  @action
+  observeReportCounter() {
+    this.getReports.perform();
+  }
+
+  @action initializeDrawer() {
+    addObserver(
```
future-pirate-king, line 215: do we require this observer ? `getReports` will anyways be called when drawer opens

future-pirate-king (changes requested):

mirage/factories/file.js
Yibaebi: Done

app/components/file/report-drawer/va-reports/index.hbs
Yibaebi: Done

app/components/file/report-drawer/va-reports/index.ts
Yibaebi: Yeah. This is required for the Websocket connection that triggers an api call to get the latest report status.

mirage/factories/file-report.js
```diff
@@ -1,15 +1,29 @@
...
+  is_generating() {
     return this.progress >= 0 && this.progress <= 99;
   },
-  isGenerated() {
+
+  is_generated() {
     return this.progress == 100;
   },
+
+  generated_on_date_time() {
```
future-pirate-king, line 20: this is already defined in model so you can get the model from store for value reference

future-pirate-king (changes requested):

mirage/factories/file-report.js
Yibaebi: When using just mirage generated models, this becomes required. I just tried it out.

future-pirate-king (approved):

PR #1220, opened 2023-04-25
update ak-loader implementation

Yibaebi:


app/components/ak-loader/index.hbs
```diff
@@ -1,39 +1,60 @@
+{{!@glint-nocheck: or, not, and}}
 <span
   data-test-ak-loader
   ...attributes
   {{style width=(concat this.size 'px') height=(concat this.size 'px')}}
-  class='ak-loader-root'
   local-class='ak-loader-root'
   role='progressbar'
 >
+  <span local-class='ak-loader-svg-label'>
```
future-pirate-king, line 9: can we just have the label block instead this would the keep the component lot more simpler

app/components/ak-loader/linear/index.hbs
```diff
@@ -0,0 +1,47 @@
...
+      {{style left=(concat this.progressIndicatorPosition '%')}}
+      data-test-ak-loader-linear-indicator
+    ></span>
+
+    {{#if this.isIndeterminateLoader}}
+      <span local-class='indeterminate-indicator-bar--2'></span>
+    {{/if}}
+  </span>
+
+  {{#unless this.hideLabel}}
```
future-pirate-king, line 30: same here too just label block

app/components/ak-button/index.scss
```diff
@@ -131,8 +131,8 @@
   align-items: center;
   margin-right: 0.5em;
 
-  :global(.ak-loader-circle-foreground) {
-    stroke: var(--ak-button-color-primary-contrast-text);
+  :global(.ak-loader-circle-indicator) {
+    stroke: var(--ak-button-color-primary-contrast-text) !important;
```
future-pirate-king, line 135: is `!important` required here ?

future-pirate-king (changes requested):

app/components/ak-loader/index.hbs
Yibaebi: It needs to be positioned absolutely. So this wrapping element is necessary.

app/components/ak-button/index.scss
Yibaebi: Yeah

app/components/ak-loader/index.hbs
future-pirate-king: has-block check can be above wrapping element and label block inside the wrapper

future-pirate-king (approved):

PR #1231, opened 2023-06-21
datepicker and radio button components refactor to typescript

avzz-19:


app/components/ak-radio/index.ts
```diff
@@ -17,7 +37,7 @@ export default class AkRadioComponent extends Component {
   }
 
   get checked() {
-    if (this.args.checked) {
+    if (this.args.checked !== undefined) {
```
future-pirate-king, line 40: use this `if (!isEmpty(this.args.checked)) {` it will check for empty string and null too

app/components/ak-radio/index.ts
```diff
@@ -2,13 +2,33 @@ import Component from '@glimmer/component';
...
+    radioCtx?: {
+      name: string;
+      value: string | number;
+      handleChange: (event: Event) => void;
+    };
+    onChange?: (event: Event, checked: boolean) => void;
+  };
+  Element: HTMLInputElement;
+  Blocks: {
+    default: [];
```
future-pirate-king, line 22: there is no block exposed so we can remove this

app/components/ak-radio/index.ts
```diff
@@ -2,13 +2,33 @@ import Component from '@glimmer/component';
...
+  };
+}
+
+export default class AkRadioComponent extends Component<AkRadioSignature> {
+  get id(): string {
     return this.args.id || `ak-radio-${guidFor(this)}`;
   }
 
-  get name() {
-    // while in group name is required for onchange to properly work
```
future-pirate-king, line 11: please restore comment

future-pirate-king (changes requested):

app/components/ak-radio/index.ts
avzz-19: Done

future-pirate-king (approved):

PR #1250, opened 2023-08-29
fix organization team invitation list resend delete

avzz-19:


app/models/organization-team-invitation.ts
```diff
@@ -1,18 +1,18 @@
...
-  @attr('string')
-  declare email: string;
+export default class OrganizationTeamInvitationModel extends OrganizationInvitationModel {
+  async delete(
+    teamInvite: OrganizationTeamInvitationModel
+  ): Promise<OrganizationTeamInvitationModel> {
+    const invite = await this.store.findRecord(
+      'organization-invitation',
+      teamInvite.id
+    );
```
future-pirate-king, line 10: add new line here

future-pirate-king (changes requested):

app/models/organization-team-invitation.ts
```diff
@@ -1,18 +1,19 @@
-import Model, { AsyncBelongsTo, attr, belongsTo } from '@ember-data/model';
-import OrganizationModel from './organization';
+import OrganizationInvitationModel from './organization-invitation';
```
cosmosgenius, line 1: This is a bad idea. Even tho the two models look similar it better to keep them separate... As in terms of business logic it is different.

Inheritance is model is bad in general case... unless you are inheriting from abstract class. Lets avoid these

cosmosgenius: Duplicate the code along with tests.

cosmosgenius (changes requested):

future-pirate-king (approved):

cosmosgenius (approved):

PR #1268, opened 2023-10-03
updates to file compare category and list sorting

Yibaebi:


app/utils/compare-files.ts
```diff
@@ -40,15 +50,14 @@ const sortByFile2Analyses = (
...
-  if (analysis2ComputedRisk > analysis1ComputedRisk) {
-    return 1;
-  }
-
-  return 0;
+  return (
+    Number(sortPriorityMap[analysis2ComputedRisk]) -
+    Number(sortPriorityMap[analysis1ComputedRisk])
+  );
 };
```
future-pirate-king, line 61: we can reduce this to a single function by having `analysis1` & `analysis2` parameterised

app/utils/compare-files.ts
```diff
@@ -40,15 +50,14 @@ const sortByFile2Analyses = (
...
     return -1;
   }
 
-  if (analysis2ComputedRisk > analysis1ComputedRisk) {
-    return 1;
-  }
-
-  return 0;
+  return (
+    Number(sortPriorityMap[analysis2ComputedRisk]) -
```
future-pirate-king, line 58: do we require number conversion since it is going to return a number

future-pirate-king (changes requested):

app/utils/compare-files.ts
Yibaebi: Yeah. Typescript safety check. Since `sortPriorityMap` is an object with keys as type `number` typescript thinks that that expression could return undefined. Even though the keys would always fall within the numbers predefined in the `sortPriorityMap` object.

future-pirate-king (approved):

PR #1269, opened 2023-10-04
platform icons color update

avzz-19:


app/styles/_global-styles.scss
```diff
@@ -5245,7 +5245,7 @@ ul.email-list {
 
 .platform-icon {
   &--android {
-    color: #7bb054;
+    color: #33A852;
```
Praseetha-KR, line 5248: This can use `var(--android)` right?

app/styles/_component-variables.scss
```diff
@@ -502,7 +502,7 @@ body {
 
   //variables for appmonitoring
   --appmonitoring-font-weight-medium: var(--font-weight-medium);
-  --appmonitoring-color-black: var(--common-black);
+  --appmonitoring-color-ios: var(--ios);
```
Praseetha-KR, line 505: To be consistent with android let's use `--appmonitoring-platform-ios-color: var(--ios);`

app/components/app-monitoring/table/store-version/index.scss
```diff
@@ -1,6 +1,6 @@
 .production-version {
   letter-spacing: -0.0286em;
-  color: var(--appmonitoring-color-black);
+  color: var(--appmonitoring-color-ios);
```
Praseetha-KR, line 3: Version color should not depend on platform.. let it be `--appmonitoring-color-black` itself

Praseetha-KR (changes requested):

PR #1289, opened 2023-10-31
Owasp Api Top 10 2023 section addition

avzz-19:


app/adapters/owaspapi2023.ts
```diff
@@ -0,0 +1,19 @@
+import CommonDRFAdapter from './commondrf';
+
+export default class OwaspApi2023Adapter extends CommonDRFAdapter {
+  _buildURL(modelName: string | number, id: string | number) {
```
cosmosgenius, line 4: don't think this override is required

app/serializers/owaspapi2023.js
```diff
@@ -0,0 +1,3 @@
+import DRFSerializer from './drf';
+
+export default DRFSerializer.extend({});
```
cosmosgenius, line 3: this file is also not required since the default is drf now

cosmosgenius (changes requested):

app/adapters/owaspapi2023.ts
```diff
@@ -0,0 +1,19 @@
...
+    const baseurl = `${this.namespace_v2}/owaspapi2023s`;
+
+    if (id) {
+      return this.buildURLFromBase(`${baseurl}/${encodeURIComponent(id)}`);
+    }
+
+    return this.buildURLFromBase(baseurl);
+  }
```
cosmosgenius, line 12:
```suggestion
  namespace = this.namespace_v2;
```

cosmosgenius: the base class should handle the rest

cosmosgenius (changes requested):

cosmosgenius (approved):

PR #1293, opened 2023-11-14
Remove ember-awesome-macros & add ember-data-resources

future-pirate-king:


app/components/password-reset/index.ts
```diff
@@ -0,0 +1,122 @@
...
+    token?: string;
+  };
+}
+
+type ChangesetBufferProps = Partial<BufferedChangeset> & {
+  password: string;
+  confirm_password: string;
+};
+
+const ResetValidator = {
```
Yibaebi, line 30: This should be moved to the validations folder.

future-pirate-king: hmm I think we should move away from this and keep validators close to the components unless it is going to be reused. if the `validator` is small then keep it in same `ts` file otherwise create a file `validator.ts` in the same component folder

Yibaebi (approved):

PR #1341, opened 2024-02-26
upgrade node & linting deps

future-pirate-king:


app/models/organization-archive.ts
```diff
@@ -7,6 +7,7 @@ import IntlService from 'ember-intl/services/intl';
 export enum OrganizationArchiveModelGeneratedVia {
   DASHBOARD = 0,
   CRM = 1,
```
Yibaebi, line 9: It's out of the context of the PR but should the `CRM` and `PARTNER` be of the same value?

Yibaebi (approved):

PR #1361, opened 2024-03-27
update cypress translations

Yibaebi:


cypress.config.ts
```diff
@@ -11,6 +11,7 @@ export default defineConfig({
       hideCredentials: true,
       TEST_USERNAME: '***',
       TEST_PASSWORD: '***',
+      TEST_GOOGLE_PASSWORD: '***',
```
future-pirate-king, line 14: why do we require this ?

cypress/support/translations.ts
```diff
@@ -1,7 +1,56 @@
...
+}
+
+/**
+ * Translates a message using the provided key, substitutes variables, and formats it based on locale.
+ * @param {NestedKeyOf<typeof EN_TRANSLATIONS>} key - The key to retrieve the message value.
+ * @param {Record<string, string | number>} [variables] - Optional variables to substitute in the message.
+ * @param {keyof typeof MESSAGES_MAP} [locale] - Optional locale to format the message. Defaults to 'en' if not provided.
+ * @returns {string} The translated and formatted message.
+ */
+export function cypressTranslate(
```
future-pirate-king, line 45: can we rename this to `cyTranslate` and can we remove `export` as it is the default export

cypress/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "lib": ["es2015", "dom", "es2018.regexp"],
     "isolatedModules": false,
     "allowJs": true,
-    "noEmit": true
+    "noEmit": true,
+    "moduleResolution": "node"
```
future-pirate-king, line 11: why is this required ?

future-pirate-king (changes requested):

cypress.config.ts
Yibaebi: It's not required for this PR. I'll take it off.

future-pirate-king (approved):

PR #1366, opened 2024-04-01
Fix:File Select Button Showing Up in Safari

SmitGala:


app/components/upload-app/via-system/index.hbs
```diff
@@ -13,9 +13,9 @@
 
     <input
       data-test-uploadApp-input
+      local-class='upload-input'
       type='file'
       id='upload-app-input'
-      {{style width='1px' height='1px'}}
```
future-pirate-king, line 18: we can just directly add `{{style display='none'}}`

future-pirate-king (changes requested):

future-pirate-king (approved):

PR #1410, opened 2024-06-27
Add support for specific devicefarm url for specific organization

krishnendu:
Use organization devicefarm url if present instead of env

### DIFF
--- a/app/routes/application.ts
+++ b/app/routes/application.ts
@@ -19,6 +19,7 @@ export default class ApplicationRoute extends Route {
     await all([
       this.configuration.serverConfigFetch(),
       this.configuration.getFrontendConfig(),
+      this.configuration.getDashboardConfig(),
     ]);
 
     return this.intl.setLocale(['en']);
--- a/app/routes/application.ts
+++ b/app/routes/application.ts
@@ -19,7 +19,6 @@ export default class ApplicationRoute extends Route {
     await all([
       this.configuration.serverConfigFetch(),
       this.configuration.getFrontendConfig(),
-      this.configuration.getDashboardConfig(),
     ]);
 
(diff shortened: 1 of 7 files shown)


app/routes/application.ts
```diff
@@ -19,6 +19,7 @@ export default class ApplicationRoute extends Route {
     await all([
       this.configuration.serverConfigFetch(),
       this.configuration.getFrontendConfig(),
+      this.configuration.getDashboardConfig(),
```
cosmosgenius, line 22: `getDashboardConfig` should be inside authenticated route as authentication is required for this.

app/services/configuration.ts
```diff
@@ -62,8 +72,34 @@ export default class ConfigurationService extends Service {
...
+      'X-Product': `${ENV.product}`,
+    };
+  }
+
+  async fetchConfig(url: string, headers?: Header) {
+    const reqOptions: RequestOptions = {};
+
+    if (headers) {
+      reqOptions.headers = headers;
+    }
```
cosmosgenius, line 100: These are not required, as network service handles authentication internally.

app/services/configuration.ts
```diff
@@ -11,12 +13,20 @@ type ServerData = {
   urlUploadAllowed: boolean;
 };
 
+type DashboardData = {
+  dashboardURL: string;
+  orgDevicefarmURL: string;
```
cosmosgenius, line 18:
```suggestion
  devicefarmURL: string;
```

cosmosgenius (changes requested):

cosmosgenius (approved):

cosmosgenius (approved):

PR #1427, opened 2024-08-06
Public API documentation integration and Customization

Yibaebi:


app/templates/authenticated/dashboard/public-api.hbs
```diff
@@ -0,0 +1,3 @@
+{{page-title 'Public Api'}}
```
future-pirate-king, line 1: use translation here

app/templates/authenticated/dashboard/public-api/docs.hbs
```diff
@@ -0,0 +1,3 @@
+{{page-title 'Docs'}}
```
future-pirate-king, line 1: use translation here

app/styles/_public-api.scss
future-pirate-king, line 1: can we move this to local styles

future-pirate-king (changes requested):

app/styles/_public-api.scss
Yibaebi: Is there any difference or extra advantage of this? 

I think third party style overrides should exist in a single file. It's also easier to maintain, update, and possibly remove styles. I was trying to stick with the same pattern for the freshdesk style overrides.

future-pirate-king: advantage is we can target the styles locally then there is no chance of interference with other parts of UI
you can still keep the styles in a single file and share the file reference or through import

future-pirate-king (approved):

PR #1440, opened 2024-08-22
Org data export based on type

future-pirate-king:


app/components/organization-archive/list/action/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export default class OrganizationArchiveListActionComponent extends Component<OrganizationArchiveListActionSignature> {
+  @service('notifications') declare notify: NotificationService;
+  @service declare intl: IntlService;
+
+  downloadArchive = task(async () => {
+    const downloadURL = await waitForPromise(this.args.archive.downloadURL());
+
+    if (downloadURL) {
+      window.open(downloadURL);
```
Yibaebi, line 23: Use window service

app/components/organization-archive/list/index.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  get columns() {
+    return [
+      {
+        name: this.intl.t('organizationTableCreatedOn'),
+        component: 'organization-archive/list/created-on',
+      },
+      {
+        name: this.intl.t('organizationTableGeneratedBy'),
+        component: 'organization-archive/list/generated-by',
```
Yibaebi, line 66: Do we need a component for `generated-by` column? We can just the `valuePath` right?

Yibaebi (changes requested):

app/components/organization-archive/list/action/index.ts
future-pirate-king: done

Yibaebi (approved):

PR #1449, opened 2024-09-04
billing page global css refactor

future-pirate-king:


app/components/organization-billing/plan/index.scss
```diff
@@ -0,0 +1,57 @@
...
+
+  .plan-text {
+    text-transform: uppercase;
+  }
+
+  .minus-icon,
+  .plus-icon {
+    color: var(--organization-billing-plan-common-white) !important;
+    border-radius: 50%;
+    font-size: 20px;
```
Yibaebi, line 32: Can we use relative units instead?

Yibaebi (changes requested):

app/components/organization-billing/plan/index.scss
future-pirate-king: We can but I button size is dependent on this which needs to be constant

PR #1505, opened 2024-12-12
ember exam test partitioning

avzz-19:


.github/workflows/test.yaml
```diff
@@ -5,16 +5,109 @@ on:
...
+
+      - name: Upload built ember app
+        uses: actions/upload-artifact@v4
+        with:
+          name: dist
+          path: dist
+
+  test:
+    name: Run ${{ matrix.tests.type }}
+    needs: [setup, lint, build]
```
future-pirate-king, line 63: test is not dependent on lint

app/components/upload-app/status/details/index.ts
```diff
@@ -110,6 +111,7 @@ export default class UploadAppStatusDetailsComponent extends Component<UploadApp
   }
 
   get createdOn() {
+    dayjs.extend(relativeTime);
```
future-pirate-king, line 114: this to top as suggested

future-pirate-king (changes requested):

PR #1518, opened 2025-01-10
Pd 1635 remove deprecated dast apis

avzz-19:


app/models/file.ts
```diff
@@ -185,151 +185,6 @@ export default class FileModel extends ModelBaseMixin {
     return this.scanProgressClass(isStaticDone);
   }
 
-  get isNoneStatus() {
-    const status = this.dynamicStatus;
```
Yibaebi, line 189: Is `dynamnicStatus` property in this model still required? I don't think it was deleted.

mirage/factories/project.ts
```diff
@@ -42,7 +42,7 @@ export const PROJECT_FACTORY_DEF = {
   },
 
   device_type() {
-    return faker.helpers.arrayElement(ENUMS.DEVICE_TYPE.VALUES);
+    return faker.helpers.arrayElement(ENUMS.DS_AUTOMATED_DEVICE_TYPE.VALUES);
```
Yibaebi, line 45: Also refer to my earlier comment about renaming `DS_AUTOMATED_DEVICE_TYPE` enum to something more generic if it is being used in both automated and manual DAST. This file and in other places it is being used.

Yibaebi (approved):

PR #1546, opened 2025-03-19
Partner dashboard template/scripts migrations

Yibaebi:


app/components/partner/client-report-summary/index.ts
```diff
@@ -0,0 +1,81 @@
+import Store from '@ember-data/store';
+import { inject as service } from '@ember/service';
```
avzz-19, line 2: we can directly import service

app/components/partner/invitation-list/index.ts
```diff
@@ -4,16 +4,16 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { inject as service } from '@ember/service';
```
avzz-19, line 5: we can directly import service

app/components/partner/client-report-summary/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    return this.riskWidthStyle(this.fileSummary?.lowPercent);
+  }
+
+  get passedWidthStyle() {
+    return this.riskWidthStyle(this.fileSummary?.passedPercent);
+  }
+
+  get untestedWidthStyle() {
+    return this.riskWidthStyle(this.fileSummary?.untestedPercent);
+  }
```
avzz-19, line 59: we can use a single dynamic getter here, and use each loop in the hbs to reduce repitition

avzz-19 (approved):

PR #1558, opened 2025-04-16
Convert MFA to Glimmer Component

SmitGala:


app/components/account-settings/security/mfa-switch-to-email/index.ts
```diff
@@ -0,0 +1,159 @@
...
+      const otpMsg = errorObj.otp && errorObj.otp[0];
+
+      if (otpMsg) {
+        this.notify.error(this.tInvalidOTP);
+      } else {
+        this.notify.error(this.tsomethingWentWrong);
+      }
+    }
+
+    return false;
```
Yibaebi, line 151: Why is there a return here?

Yibaebi (approved):

PR #1567, opened 2025-05-23
integrations page revamp and splunk integration

avzz-19:


app/components/project-settings/integrations/splunk/index.ts
```diff
@@ -0,0 +1,270 @@
...
+      }
+
+      const error = err as AdapterError;
+      const { source, detail: errorDetail } = error?.errors?.[0] || {};
+
+      if (
+        [
+          '/data/attributes/riskThreshold',
+          '/data/attributes/risk_threshold',
+        ].indexOf(String(source?.pointer)) > -1
```
Yibaebi, line 248: We can use a single assertion 

```
      [
          '/data/attributes/riskThreshold',
          '/data/attributes/risk_threshold',
        ].includes(String(source?.pointer))
 ```

File: tests/integration/components/project-settings/integrations/github-project-test.js [hunk 2]
```diff
@@ -1,5 +1,5 @@
 import Service from '@ember/service';
```
Yibaebi, line 1: On this file there are many instances of `assert.dom().exists().someAssertion()` we can rid off the `.exists()` assertions since every assertion after it would fail if the element being asserted on doesn't exist. This would reduce the number off assertions being made on this test module.

translations/en.json
```diff
@@ -665,7 +676,7 @@
   "jiraHost": "Jira Host",
   "jiraIntegrated": "JIRA integrated",
   "jiraIntegration": "JIRA Integration",
-  "jiraIntegrationDesc": "Here you can intergrate your JIRA Account",
+  "jiraIntegrationDesc": "It’s project management tool which helps teams plan, track, and manage work, especially in software development.",
```
Yibaebi, line 679: `It's project management ...` should be `It's a project management ...`.

translations/ja.json
```diff
@@ -665,7 +676,7 @@
   "jiraHost": "JIRAホスト",
   "jiraIntegrated": "JIRAと連携されました",
   "jiraIntegration": "JIRA連携",
-  "jiraIntegrationDesc": "Here you can intergrate your JIRA Account",
+  "jiraIntegrationDesc": "It’s project management tool which helps teams plan, track, and manage work, especially in software development.",
```
Yibaebi, line 679: We can update this as well.

Yibaebi (changes requested):

app/components/organization/integrations/splunk/index.ts
```diff
@@ -0,0 +1,441 @@
...
+        .slice(0)
+        .reverse()
+        .map((index) => ({
+          label: index,
+          value: index,
+        })) as indexItem[];
+
+      this.indexOptions = indexes;
+
+      if (indexes.length > 0) {
```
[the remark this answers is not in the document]
Yibaebi: Seems like the logic here was updated. If so, we can simplify to

```
      const indexes = data.indexes
        .filter((index) => !index.startsWith('_'))
        .reverse()
        .map((index) => ({ label: index, value: index })) as indexItem[];

      this.indexOptions = indexes;

      if (indexes.length > 0) {
        const currVulnIndex = this.selectedVulnIndex?.value;
        const currUserLogIndex = this.selectedUserLogIndex?.value;

        // If the current vuln index is not in the indexes array, set the first index as the selected vuln index
        const selectedVulnIndex =
          indexes.find((idx) => idx.value === currVulnIndex) ||
          this.indexOptions[0];

        this.setVulnIndex(selectedVulnIndex as indexItem);

        // If the current user log index is not in the indexes array, set the second index as the selected user log index
        const selectedUserLogIndex =
          indexes.find((idx) => idx.value === currUserLogIndex) ||
          this.indexOptions[1] ||
          this.indexOptions[0];

        this.setUserLogIndex(selectedUserLogIndex as indexItem);
      }

      return indexes;
```

Too many nested if statements make the code less readable. I've also removed the `.slice()` methods as the `.filter()` would create the required array.

app/components/project-settings/integrations/splunk/index.ts
```diff
@@ -0,0 +1,270 @@
...
+      this.splunkConfig = config;
+      this.selectedThreshold = config.riskThreshold;
+    } catch (err) {
+      const error = err as AdapterError;
+      const errorInfo = error.errors?.[0];
+      const errorDetail = errorInfo?.detail;
+
+      if (errorDetail === 'Splunk not integrated') {
+        this.noIntegration = true;
+        return;
```
Yibaebi, line 161: Add spaces before return in this code block.

Yibaebi (changes requested):

Yibaebi (approved):

PR #1568, opened 2025-05-27
PII - Privacy Module

SmitGala:


tests/integration/components/privacy/pii-test.js
```diff
@@ -0,0 +1,359 @@
...
+
+    await click(allPiiCategory[0]);
+
+    const allDataFound = findAll(
+      '[data-test-privacyModule-pii-drawer-data-found]'
+    );
+
+    assert
+      .dom('[data-test-privacyModule-pii-drawer-header]')
+      .exists()
```
Yibaebi, line 184: We can remove `.exists()` assertions if is is chained with other assertions besides the `.dom()` assertion. This should be `assert.dom().hasText()`. This reduces the number of assertions we produce over time.

Yibaebi (approved):

Yibaebi (approved):

PR #1605, opened 2025-09-04
Remove Pagination from PII

SmitGala:


app/services/privacy-module.ts
```diff
@@ -146,67 +142,64 @@ export default class PrivacyModuleService extends Service {
...
-        ) {
-          const queryParams = {
-            limit: limit,
-            offset: offset,
-            fileId: fileId,
-            piiExtractionId: pii.id,
-          };
+        const piiDataList = (
+          await this.store.query('pii', queryParams)
+        ).toArray() as PiiModel[];
```
Yibaebi, line 162: Use `.slice` method instead and separate query call from the .slice conversion.

Yibaebi (changes requested):

PR #1683, opened 2026-04-28
Improve: the DAST device and VNC experience

sammyjeng:
## Summary

  This PR refines the dynamic scan device/VNC flow with support for newer iOS device behavior, more reliable realtime scan updates, and stronger test coverage around the updated
  socket path.

  ## Changes

  - Added iOS device version helpers for manual dynamic scan device selection.
    - Legacy iOS apps continue to request devices below iOS 17.
    - iOS 17+ apps request modern iOS devices using the app minimum OS version.

  - Updated the VNC viewer to support the modern iOS device frame.
    - Modern iOS devices render in the new iOS frame.
    - Legacy iOS and Android devices keep the existing `marvel-device` frame path.
    - noVNC password is skipped for modern iOS device frames and retained for the existing VNC path.

  - Hardened the noVNC RFB lifecycle.
    - Handles missing/invalid VNC URLs safely.
    - Makes the `connect` listener one-shot after successful handshake.
    - Explicitly removes the listener during teardown because `RFB.disconnect()` does not clear user-added EventTarget listeners.

  - Routed dynamic scan websocket updates through the event bus.
    - `model_updated` notifications now push the updated `dynamicscan` into the store and trigger `ws:dynamicscan:update`.
    - Manual and automated scan views reload their latest scan state from that event.

  - Improved manual device selection behavior.
    - The drawer tracks whether the currently selected specific device is available in the fetched device list.
    - Start remains disabled when a specific device is selected but unavailable.

  ## Tests

  - Added/updated integration coverage for:
    - iOS manual device version filtering.
    - Specific manual device selection and start-button enablement.
    - Manual scan realtime reload from websocket/event-bus updates.
    - Modern iOS VNC frame rendering.

  - Updated acceptance coverage to simulate the new `model_updated` websocket notification path instead of the old object event path.

### DIFF
--- a/app/components/file-details/dynamic-scan/action/drawer/device-pref-table/index.ts
+++ b/app/components/file-details/dynamic-scan/action/drawer/device-pref-table/index.ts
@@ -11,6 +11,7 @@ import type DS from 'ember-data';
 
 import styles from './index.scss';
 import ENUMS from 'irene/enums';
+import { getIOSManualDeviceVersionParams } from 'irene/utils/dynamic-scan-device';
 import { type PaginationProviderActionsArgs } from 'irene/components/ak-pagination-provider';
 import type { DsPreferenceContext } from 'irene/components/ds-preference-provider';
 import type FileModel from 'irene/models/file';
@@ -50,6 +51,7 @@ export interface FileDetailsDynamicScanDrawerDevicePrefTableSignature {
   Args: {
     dpContext: DsPreferenceContext;
     file: FileModel;
+    onSelectedManualDeviceAvailable(deviceIdentifier: string): void;
   };
 }
 
@@ -185,6 +187,8 @@ export default class FileDetailsDynamicScanDrawerDevicePrefTableComponent extend
 
     preference.dsManualDeviceIdentifier = device.deviceIdentifier;
 
+    this.args.onSelectedManualDeviceAvailable(device.deviceIdentifier);
+
     this.dpContext.updateDsManualDevicePref(preference);
   }
 
@@ -230,8 +234,25 @@ export default class FileDetailsDynamicScanDrawerDevicePrefTableComponent extend
 
         this.availableDevicesResponse = (await this.store.query(
           'available-manual-device',
-          { ...queryParams, platform_version_min: this.args.file.minOsVersion }
+          {
+            ...queryParams,
+            ...getIOSManualDeviceVersionParams(this.args.file),
+          }
         )) as AvailableManualDeviceQueryResponse;
+
+        const selectedDeviceIdentifier =
+          this.devicePreference?.dsManualDeviceIdentifier;
+
+        const selectedDeviceIsAvailableInResponse =
+          this.availableManualDevices.some(
+            (device) => device.deviceIdentifier === selectedDeviceIdentifier
+          );
+
+        if (selectedDeviceIdentifier && selectedDeviceIsAvailableInResponse) {
+          this.args.onSelectedManualDeviceAvailable(selectedDeviceIdentifier);
+        } else {
+          this.args.onSelectedManualDeviceAvailable('');
+        }
       } catch (error) {
         const err = error as AdapterError;
         const errorStatus = err.errors?.[0]?.status;
--- a/app/components/file-details/dynamic-scan/action/drawer/index.ts
+++ b/app/components/file-details/dynamic-scan/action/drawer/index.ts
@@ -8,6 +8,7 @@ import { isEmpty } from '@ember/utils';
 import ENV from 'irene/config/environment';
 import ENUMS from 'irene/enums';
 import parseError from 'irene/utils/parse-error';
+import { getIOSManualDeviceVersionParams } from 'irene/utils/dynamic-scan-device';
 
 import type IntlService from 'ember-intl/services/intl';
 import type Store from 'ember-data/store';
@@ -39,6 +40,7 @@ export default class FileDetailsDynamicScanActionDrawerComponent extends Compone
 
   @tracked isApiCaptureEnabled = false;
   @tracked availableManualDevices: AvailableManualDeviceModel[] = [];
+  @tracked selectedAvailableManualDeviceIdentifier: string | null = null;
 
   constructor(
     owner: unknown,
@@ -80,6 +82,13 @@ export default class FileDetailsDynamicScanActionDrawerComponent extends Compone
     );
   }
 
+  get selectedManualDeviceIsFromAvailableDeviceTable() {
+    return (
+      this.selectedAvailableManualDeviceIdentifier ===
+      this.dsManualDeviceIdentifier
+    );
+  }
+
   get enableStartDynamicScanBtn() {
     const { isAutomatedScan, dpContext } = this.args;
 
@@ -96,11 +105,17 @@ export default class FileDetailsDynamicScanActionDrawerComponent extends Compone
         dsManualDeviceSelection === anyDeviceSelection ||
         (specificDeviceSelection === dsManualDeviceSelection &&
           !isEmpty(this.dsManualDeviceIdentifier) &&
-          this.selectedManualDeviceIsInAvailableDeviceList)
+          (this.selectedManualDeviceIsInAvailableDeviceList ||
+            this.selectedManualDeviceIsFromAvailableDeviceTable))
       );
     }
 
     return true;
+  }
+
+  @action
+  handleSelectedManualDeviceAvailable(deviceIdentifier: string) {
+    this.selectedAvailableManualDeviceIdentifier = deviceIdentifier || null;
   }
 
   @action
@@ -122,10 +137,21 @@ export default class FileDetailsDynamicScanActionDrawerComponent extends Compone
       );
 
       const devices = await this.store.query('available-manual-device', {
-        platform_version_min: this.args.file.minOsVersion,
+        ...getIOSManualDeviceVersionParams(this.file),
       });
 
       this.availableManualDevices = devices.slice();
+
+      if (
+        this.selectedAvailableManualDeviceIdentifier &&
+        !this.availableManualDevices.some(
+          (device) =>
+            device.deviceIdentifier ===
+            this.selectedAvailableManualDeviceIdentifier
+        )
+      ) {
+        this.selectedAvailableManualDeviceIdentifier = null;
+      }
     } catch (error) {
       const err = error as AdapterError;
       const errorStatus = err.errors?.[0]?.status;
--- a/app/components/file-details/dynamic-scan/automated/index.ts
+++ b/app/components/file-details/dynamic-scan/automated/index.ts
@@ -13,6 +13,7 @@ import type DsAutomationPreferenceModel from 'irene/models/ds-automation-prefere
 import type OrganizationService from 'irene/services/organization';
 import type DynamicscanModel from 'irene/models/dynamicscan';
 import type LoggerService from 'irene/services/logger';
+import type EventBusService from 'irene/services/event-bus';
 
 export interface FileDetailsDastAutomatedSignature {
   Args: {
@@ -28,6 +29,7 @@ export default class FileDetailsDastAutomated extends Component<FileDetailsDastA
   @service declare organization: OrganizationService;
   @service('notifications') declare notify: NotificationService;
   @service declare logger: LoggerService;
+  @service declare eventBus: EventBusService;
 
   @tracked automationPreference: DsAutomationPreferenceModel | null = null;
   @tracked lastAutomatedDynamicScan: DynamicscanModel | null = null;
@@ -37,6 +39,12 @@ export default class FileDetailsDastAutomated extends Component<FileDetailsDastA
 
     this.getDsAutomationPreference.perform();
     this.getLastAutomatedDynamicScan.perform();
+
+    this.eventBus.on(
+      'ws:dynamicscan:update',
+      this,
+      this.handleDynamicScanUpdate
+    );
   }
 
   get file() {
@@ -82,7 +90,14 @@ export default class FileDetailsDastAutomated extends Component<FileDetailsDastA
     this.getLastAutomatedDynamicScan.perform();
   }
 
-  getLastAutomatedDynamicScan = task(async () => {
+  @action
+  handleDynamicScanUpdate(dynamicscan: DynamicscanModel) {
+    if (String(dynamicscan.file.get('id')) === String(this.file.id)) {
+      this.getLastAutomatedDynamicScan.perform();
+    }
+  }
+
+  getLastAutomatedDynamicScan = task({ restartable: true }, async () => {
     try {
       this.lastAutomatedDynamicScan =
         await this.file.getFileLastAutomatedDynamicScan();
@@ -90,6 +105,16 @@ export default class FileDetailsDastAutomated extends Component<FileDetailsDastA
       this.logger.error(parseError(error, this.intl.t('pleaseTryAgain')));
     }
   });
+
+  willDestroy(): void {
+    super.willDestroy();
+
+    this.eventBus.off(
+      'ws:dynamicscan:update',
+      this,
+      this.handleDynamicScanUpdate
+    );
+  }
 }
 
 declare module '@glint/environment-ember-loose/registry' {
--- a/app/components/file-details/dynamic-scan/manual/index.ts
+++ b/app/components/file-details/dynamic-scan/manual/index.ts
@@ -9,6 +9,7 @@ import type IntlService from 'ember-intl/services/intl';
 import type FileModel from 'irene/models/file';
 import type DynamicScanService from 'irene/services/dynamic-scan';
 import type DynamicscanModel from 'irene/models/dynamicscan';
+import type EventBusService from 'irene/services/event-bus';
 
 export interface FileDetailsDastManualSignature {
   Args: {
@@ -21,6 +22,7 @@ export default class FileDetailsDastManual extends Component<FileDetailsDastManu
   @service('dynamic-scan') declare dsService: DynamicScanService;
   @service declare intl: IntlService;
   @service('notifications') declare notify: NotificationService;
+  @service declare eventBus: EventBusService;
 
   @tracked lastManualDynamicScan: DynamicscanModel | null = null;
 
@@ -35,6 +37,12 @@ export default class FileDetailsDastManual extends Component<FileDetailsDastManu
     });
 
     this.getLastDynamicScans.perform();
+
+    this.eventBus.on(
+      'ws:dynamicscan:update',
+      this,
+      this.handleDynamicScanUpdate
+    );
   }
 
   get file() {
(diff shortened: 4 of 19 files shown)


app/utils/ws-model-ev-handlers.ts
```diff
@@ -62,7 +62,9 @@ export class DynamicScanEventHandler extends WsModelEventHandler<DynamicscanMode
     this.incrementFileAutoDynamicScanReloadCounter();
   }
 
-  onUpdate() {}
+  onUpdate() {
+    this.incrementFileAutoDynamicScanReloadCounter();
```
Yibaebi: Let’s not increment the reload counter (`FileAutoDynamicScanReloadCounter`) in the onUpdate handler as this might lead to unplanned side effects. The `FileAutoDynamicScanReloadCounter` is usually incremented whenever a new dynamic scan model is created for the first time. Doing this in the `onUpdate` means it will be fired every time a pre-existing dynamic scan model changes as communicated by the backend through Websocket. 

Let us do something similar to the `class AnalysisEventHandler` and create a new event bus to trigger similar to "`ws:analysis-overview:update`" that sends events across to every location in the application. 

REFER --> `app/components/file-compare/index.ts#L58` on how to use the event bus pattern.

In summary the pattern from `AnalysisEventHandler` is:

- In onUpdate, push the incoming wsData into the store as a dynamicscan model (same as pushAnalysisOverviewAndNotify does for analysis-overview)
- Trigger a **"ws:dynamicscan:update event"** on the event bus with the updated model
- Consumers that need to react to an update (e.g. refreshing scan status UI) subscribe via `eventBus.on(...)` / `eventBus.off(...)` in `constructor` / `willDestroy`, just like `FileCompareComponent` does at line `57–61` and `180–184`

app/utils/platform-version.ts
```diff
@@ -0,0 +1,5 @@
+export function getPlatformMajorVersion(platformVersion?: string | null) {
```
Yibaebi: - We can move this utility function to (`app/utils/dynamic-scan-device-version-query.ts`) since it's only used in one place. We can delete the module after.
- Also, lets use `RegExp.exec()` in place of `.match()` as it is getting flagged on SonarQube. The implementation can be as follows.

> `const majorVersion = /\d+/.exec(platformVersion ?? '')?.[0];`

Yibaebi: Please also confirm the correctness of the implementation.

app/components/file-details/dynamic-scan/action/drawer/device-pref-table/index.ts
```diff
@@ -149,15 +153,21 @@ export default class FileDetailsDynamicScanDrawerDevicePrefTableComponent extend
...
+          isAvailableManualDeviceAllowedForFile(this.args.file, device)
+        ) || []
+    );
   }
 
   get hasNoAvailableManualDevice() {
     return this.totalAvailableManualDevicesCount === 0;
   }
 
   get totalAvailableManualDevicesCount() {
```
Yibaebi, line 161: This might break pagination of the available devices as the filter is now happening only on the frontend. Can we add a TODO comment to update this when filtering is now supported at the backend?

app/components/file-details/dynamic-scan/action/drawer/device-pref-table/index.ts
```diff
@@ -230,8 +242,23 @@ export default class FileDetailsDynamicScanDrawerDevicePrefTableComponent extend
 
         this.availableDevicesResponse = (await this.store.query(
           'available-manual-device',
-          { ...queryParams, platform_version_min: this.args.file.minOsVersion }
+          {
+            ...queryParams,
+            ...getDynamicScanDeviceVersionQuery(this.args.file),
```
Yibaebi: This function `getDynamicScanDeviceVersionQuery` can be replaced in line by:

```
this.availableDevicesResponse = (await this.store.query(
  'available-manual-device',
  {
    ...queryParams,
    ...(this.args.file.minOsVersion
      ? { platform_version_min: this.args.file.minOsVersion }
      : {}),
  }
)) as AvailableManualDeviceQueryResponse;

```

We can discard the utility to reduce the overhead at first glance.

app/components/file-details/dynamic-scan/action/drawer/index.ts
```diff
@@ -122,10 +140,14 @@ export default class FileDetailsDynamicScanActionDrawerComponent extends Compone
       );
 
       const devices = await this.store.query('available-manual-device', {
-        platform_version_min: this.args.file.minOsVersion,
+        ...getDynamicScanDeviceVersionQuery(this.args.file),
```
Yibaebi: This function getDynamicScanDeviceVersionQuery can be replaced in line by:

this.availableDevicesResponse = (await this.store.query(
  'available-manual-device',
  {
    ...queryParams,
    ...(this.args.file.minOsVersion
      ? { platform_version_min: this.args.file.minOsVersion }
      : {}),
  }
)) as AvailableManualDeviceQueryResponse;

We can discard the utility to reduce the overhead at first glance.

app/components/file-details/dynamic-scan/automated/index.ts
```diff
@@ -37,6 +40,14 @@ export default class FileDetailsDastAutomated extends Component<FileDetailsDastA
 
     this.getDsAutomationPreference.perform();
     this.getLastAutomatedDynamicScan.perform();
+
+    // eslint-disable-next-line ember/no-observers
+    addObserver(
```
Yibaebi: Please refer to the comment here: https://github.com/appknox/irene/pull/1683/changes#r3155503850

app/components/file-details/dynamic-scan/header/index.ts
```diff
@@ -55,12 +55,12 @@ export default class FileDetailsDastHeader extends Component<FileDetailsDastHead
...
 
-    // This observer is used to reload the last automated dynamic scan when the FileAutoDynamicScanReloadCounter is incremented
+    // This observer is used to reload the latest dynamic scans when the FileAutoDynamicScanReloadCounter is incremented
     // REFER: app/services/websocket.ts (LINE: 197)
     // The workflow is =>
-    // AUTO DYNAMIC SCAN STARTED -> NEW DYNAMIC SCAN CREATED ->
+    // DYNAMIC SCAN STARTED/UPDATED ->
     // "model_created" notification is sent by backend -> FileAutoDynamicScanReloadCounter is incremented on the client from the ws service
-    // -> reloadLastDynamicScans is called -> lastAutomatedDynamicScan is reloaded through this observer
+    // -> reloadLastDynamicScans is called -> latest scans are reloaded through this observer
```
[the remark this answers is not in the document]
Yibaebi: Please refer to the comment here: https://github.com/appknox/irene/pull/1683/changes#r3155503850

No changes should be on this file if implemented.

app/components/file-details/dynamic-scan/manual/index.ts
```diff
@@ -35,6 +38,14 @@ export default class FileDetailsDastManual extends Component<FileDetailsDastManu
     });
 
     this.getLastDynamicScans.perform();
+
+    // eslint-disable-next-line ember/no-observers
+    addObserver(
```
Yibaebi: Please refer to the comment here: https://github.com/appknox/irene/pull/1683/changes#r3155503850

app/components/novnc-rfb/index.ts
```diff
@@ -4,46 +4,71 @@ import { tracked } from '@glimmer/tracking';
...
+
+      rfb._display.autoscale(element.offsetWidth, element.offsetHeight);
     });
   }
 
   @action
   onWillDestroyElement() {
-    this.rfb?.removeEventListener('connect', () => {
-      this.rfb?.disconnect();
+    if (this.rfb) {
```
Yibaebi: Why was the logic here changed? Does `this.rfb?.disconnect()` removes the `this.rfb?.removeEventListener('connect')` event listener internally?

app/components/vnc-viewer/index.scss
```diff
@@ -1,3 +1,21 @@
+.ios-device-frame {
```
Yibaebi: Let's move the new device styles to `app/styles/_devices.scss` and use `class` instead of `local-class `in the template.

app/components/vnc-viewer/index.ts
```diff
@@ -8,6 +8,25 @@ import ENV from 'irene/config/environment';
...
+
+export function requiresVncPassword(platformVersion?: string | null) {
+  const majorVersion = getPlatformMajorVersion(platformVersion);
+
+  return majorVersion !== null
+    ? majorVersion < VNC_PASSWORD_AUTH_VERSION_CUTOFF
+    : false;
+}
+
+export function supportsModernIOSDeviceFrame(platformVersion?: string | null) {
```
Yibaebi: - Let's move this as a getter in the template class below
- We can also simplify to:

> ```
> get supportsModernIOSDeviceFrame() {
>   const majorVersion = getPlatformMajorVersion(this.deviceUsed?.platform_version);
> 
>   return majorVersion !== null && majorVersion >= VNC_PASSWORD_AUTH_VERSION_CUTOFF;
> }
> 
> get usesModernIOSDeviceFrame() {
>   return this.isIOSDevice && this.supportsModernIOSDeviceFrame;
> }
> ```
>

app/components/vnc-viewer/index.ts
```diff
@@ -8,6 +8,25 @@ import ENV from 'irene/config/environment';
 import type FileModel from 'irene/models/file';
 import type DynamicscanModel from 'irene/models/dynamicscan';
 import type DevicefarmService from 'irene/services/devicefarm';
+import { getPlatformMajorVersion } from 'irene/utils/platform-version';
+
+const VNC_PASSWORD_AUTH_VERSION_CUTOFF = 17;
+
+export function requiresVncPassword(platformVersion?: string | null) {
```
Yibaebi: - Let's move this as a getter in the template class below
- We can also simplify to:

> ```
> get requiresVncPassword() {
>   const majorVersion = getPlatformMajorVersion(this.deviceUsed?.platform_version);
> 
>   return majorVersion !== null && majorVersion < VNC_PASSWORD_AUTH_VERSION_CUTOFF;
> }
> 
> get deviceFarmPassword() {
>   return this.requiresVncPassword ? ENV.deviceFarmPassword : null;
> }
> 
> ```

app/components/vnc-viewer/index.ts
```diff
@@ -8,6 +8,25 @@ import ENV from 'irene/config/environment';
 import type FileModel from 'irene/models/file';
 import type DynamicscanModel from 'irene/models/dynamicscan';
 import type DevicefarmService from 'irene/services/devicefarm';
+import { getPlatformMajorVersion } from 'irene/utils/platform-version';
```
Yibaebi: We can update import here once we migrate to `app/utils/dynamic-scan-device.ts`

app/utils/dynamic-scan-device.ts
```diff
@@ -0,0 +1,54 @@
+import ENUMS from 'irene/enums';
```
Yibaebi: Can we rename to `dynamic-scan-device.ts` and move utilities from [platform-version.ts](https://github.com/appknox/irene/pull/1683/changes#diff-c80cb63bac50ff29fa44d589440317047cb44e516a7e0375700f4a6ab3f99448) to this file?

app/services/realtime.ts
```diff
@@ -26,6 +26,6 @@ export default class RealtimeService extends Service {
   @tracked InterimReportCounter = 0;
 
   // Client-side counters
-  // "FileAutoDynamicScanReloadCounter" is used to trigger a reload of the file's last automated dynamic scan in the dynamic scan header component
+  // "FileAutoDynamicScanReloadCounter" is used to trigger a reload of the file's latest dynamic scans in dynamic scan page components
```
Yibaebi: Not expecting a change in this file if https://github.com/appknox/irene/pull/1683/changes#r3155503850 comment is resolved.

app/utils/dynamic-scan-device-version-query.ts
```diff
@@ -0,0 +1,54 @@
...
+import type AvailableManualDeviceModel from 'irene/models/available-manual-device';
+import { getPlatformMajorVersion } from 'irene/utils/platform-version';
+
+const IOS_MODERN_DEVICE_VERSION_CUTOFF = 17;
+
+export interface DynamicScanDeviceVersionQuery {
+  platform_version_min?: string;
+}
+
+export function getDynamicScanDeviceVersionQuery(
```
Yibaebi: We can use the inline version instead of a separate function for this query.

tests/integration/components/file-details/dynamic-scan/manual-test.js
```diff
@@ -486,6 +489,217 @@ module(
...
+
+        await click('[data-test-fileDetails-dynamicScanAction="startBtn"]');
+
+        assert.strictEqual(
+          drawerQueryParams.platform_version_min,
+          minOsVersion
+        );
+
+        assert.strictEqual(tableQueryParams.platform_version_min, minOsVersion);
+        assert.false('platform_version_max' in drawerQueryParams);
```
Yibaebi: > TypeError can be thrown as this operand might have primitive type.

Please review this error by SonarQube. It is failing our quality checks. Can we assert in a different way?

tests/integration/components/file-details/dynamic-scan/manual-test.js
```diff
@@ -486,6 +489,217 @@ module(
...
+        assert.false('platform_version_max' in tableQueryParams);
```
Yibaebi: > TypeError can be thrown as this operand might have primitive type.

Please review this error by SonarQube. It is failing our quality checks. Can we assert in a different way?

tests/integration/components/file-details/dynamic-scan/manual-test.js
```diff
@@ -486,6 +489,217 @@ module(
...
+        this.server.get(
+          '/v2/projects/:id/available_manual_devices',
+          (schema, req) => {
+            if ('limit' in req.queryParams) {
+              tableQueryParams = req.queryParams;
+            } else {
+              drawerQueryParams = req.queryParams;
+            }
+
+            const results =
```
Yibaebi: Shouldn't we return all the devices for all cases from the API and test whether the `minOsVersion` filter works? To simulate exactly what we have implemented? I believe the filtering is happening at the client side.

tests/integration/components/file-details/dynamic-scan/manual-test.js
```diff
@@ -486,6 +489,217 @@ module(
...
+      this.server.create('dynamicscan', {
+        file: this.file.id,
+        mode: ENUMS.DYNAMIC_MODE.MANUAL,
+        status: ENUMS.DYNAMIC_SCAN_STATUS.IN_QUEUE,
+        ended_on: null,
+      });
+
+      const realtime = this.owner.lookup('service:realtime');
+
+      realtime.incrementProperty('FileAutoDynamicScanReloadCounter');
```
Yibaebi: Let's update this test based on comment --> https://github.com/appknox/irene/pull/1683/changes#r3155503850

tests/integration/components/vnc-viewer-test.js
```diff
@@ -36,6 +40,18 @@ module('Integration | Component | vnc-viewer', function (hooks) {
     });
   });
 
+  test('it requires vnc password for device platform versions below 17', function (assert) {
```
Yibaebi: We might not require these tests if we move to the component getters instead. 

We can assert this in a different way or remove entirely If not required.

tests/integration/components/vnc-viewer-test.js
```diff
@@ -6,6 +6,10 @@ import { setupIntl } from 'ember-intl/test-support';
 import { module, test } from 'qunit';
 
 import ENUMS from 'irene/enums';
+import {
+  requiresVncPassword,
```
Yibaebi: We can recreate these functions in this test if we need to use the logic in this test file. Ideally, they should be component getters not standalone functions.

We can assert the UI changes their return value causes to the UI by changing the `platform_version` property in `deviceUsed` key of the `DynamicScan` model.

Yibaebi (changes requested):

app/components/file-details/dynamic-scan/action/drawer/device-pref-table/index.ts
sammyjeng: Todo has been added for the future reference. Tested the pagination part and it is working fine.

app/utils/platform-version.ts
sammyjeng: func is removed from `app/utils/platform-version.ts` and moved to  -> `app/utils/dynamic-scan-device.ts`

tests/integration/components/file-details/dynamic-scan/manual-test.js
sammyjeng: logic moved from a front-end func that does post filtering to applying the filter via uri params -> tests covered for that as well

app/components/novnc-rfb/index.ts
sammyjeng: 1. Self-removing connect handler (novnc-rfb/index.ts:56-65): Once the handshake fires it once, the listener removes itself from noVNC's _listeners Map. The instance still tracks
  connectHandler so willDestroy can remove it for the rare case where teardown beats the handshake.
2. Comments document the noVNC contract: One comment on the listener (one-shot intent + why we still track it on the instance), one in willDestroy answering the reviewer's exact
  question — RFB.disconnect() only tears down its internal socket listeners (_sock.off), it does not clear listeners registered via the user-facing EventTargetMixin.addEventListener.
 So we explicitly remove ours.

avzz-19 (approved):
LGTM

app/components/novnc-rfb/index.scss
```diff
@@ -6,4 +6,16 @@
     cursor: pointer !important;
     overflow: hidden;
   }
+
+  &[data-contain-canvas='true'] {
```
future-pirate-king, line 10: any specific reason we are using this data attribute instead of a class ?

app/components/vnc-viewer/index.hbs
```diff
@@ -58,61 +58,100 @@
...
-      <div data-test-vncViewer-deviceTopBar class='top-bar' />
-      <div data-test-vncViewer-deviceSleep class='sleep' />
-      <div data-test-vncViewer-deviceVolume class='volume' />
-    {{/if}}
-
-    <div data-test-vncViewer-deviceCamera class='camera' />
-
-    {{#if this.isIOSDevice}}
+  {{#if this.supportsModernIOSDeviceFrame}}
+    <div data-test-vncViewer-device class='ios-device-frame'>
```
future-pirate-king, line 62: can we deduplicate or create a new component individual device frame

future-pirate-king (changes requested):

PR #1692, opened 2026-06-22
Knoxiq: All False Positive Findings

SmitGala:


app/enums.ts
```diff
@@ -507,7 +507,7 @@ const ENUMS = {
   },
 
   KNOXIQ_EXPLOITABILITY: {
-    UNKNOWN: 0,
+    UNKWN: 0,
```
Yibaebi: Instead of calling this `UNKWN` lets use another term like `EXP_UNKNOWN` its is more understandable than `UNKWN`.

Yibaebi (changes requested):

Yibaebi (approved):

PR #1712, opened 2026-08-10
feat(cyod): models, enums and shared UI foundations

ssingh55:
1/7 in the CYOD stack. Bases on `develop`. Merge in order 01→07.

Foundations the rest of the stack builds on: device/organization model fields, CYOD enums, the `cyod-settings` route, the `isCyodEnabled`/`isCyodRegistrationEnabled` gates on the organization service, and the shared `resolveFileProject` helper. Also carries the full translation set for the stack and an `@fontWeight` passthrough on `AkTextField`.

Translations land whole here rather than being split per PR — a missed key is a runtime failure, not a build error.

cc @Yibaebi — design review across the stack; the visible surfaces start in 02.


app/components/organization/settings-panel-header/index.hbs
```diff
@@ -0,0 +1,43 @@
+<AkStack
+  local-class='panel-header'
```
SmitGala: no need of this class

use @witdh='full'

app/components/organization/settings-panel-header/index.hbs
```diff
@@ -0,0 +1,43 @@
...
+  @justifyContent='space-between'
+  ...attributes
+>
+  <AkStack @spacing='1.5' @alignItems='center'>
+    {{#if @iconName}}
+      <div
+        local-class='panel-icon'
+        data-test-orgSettingsPanelHeader-icon
+        aria-hidden='true'
+      >
```
SmitGala: Use AkStack instead ...will save from custom css flex and aligning

translations/en.json
SmitGala, line 1: I see translations added are already used in this file

either use same or make an object different for cyod itself

cyod: { //all the translations 
}

some of them are repeated in this changes itslef... eg. "Active"

app/components/organization/settings-panel-header/index.scss
```diff
@@ -0,0 +1,32 @@
+.panel-header {
+  width: 100%;
+}
```
SmitGala: Remove this

app/components/organization/settings-panel-header/index.scss
```diff
@@ -0,0 +1,32 @@
...
+
+.panel-icon {
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
SmitGala: Remove after hbs changes

SmitGala (changes requested):

PR #1713, opened 2026-08-10
feat(cyod): account settings CYOD page and device registration

ssingh55:
2/7 in the CYOD stack. Bases on #1712 (cyod-01-foundation). Merge in order 01→07.

The Account Settings → CYOD Settings tab: the register-a-device drawer with its Mercer setup steps, the device table, and the two empty states (no devices registered / CYOD turned off).

Design review applied: 500px drawer, DM Mono server URL at 12px with an 11px token hint, 14px/600 empty-state headers over 13px subtext capped at 345px, and the tray illustration from the Figma frame.

cc @Yibaebi

PR #1714, opened 2026-08-10
feat(cyod): owner switch for organization device registration

ssingh55:
3/7 in the CYOD stack. Bases on #1713 (cyod-02-account-settings). Merge in order 01→07.

The owner-facing CYOD Registration switch and its device table. Layered under the paid `Feature.cyod` entitlement: the entitlement decides whether an org may use CYOD at all, this decides whether members may currently register their own devices.

Design review applied: title at h5 (16px/700) sharing a centre-aligned row with its toggle, subtext capped at 500px, status chips centred at medium weight.

The component ships here; it is wired into the settings page in 04, which is where the acceptance test lives.

cc @Yibaebi

### DIFF
--- a/app/components/organization/device-registration/index.hbs
+++ b/app/components/organization/device-registration/index.hbs
@@ -1,0 +1,57 @@
+<div class='w-full' data-test-orgDeviceRegistration ...attributes>
+  <AkStack @direction='column' @spacing='0.5'>
+    <AkStack
+      @width='full'
+      @alignItems='center'
+      @justifyContent='space-between'
+      @spacing='2'
+    >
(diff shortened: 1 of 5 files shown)


app/components/organization/device-registration/index.hbs
```diff
@@ -0,0 +1,63 @@
+<div
+  local-class='device-registration-container'
```
SmitGala: add class='w-full'

app/components/organization/device-registration/index.scss
```diff
@@ -0,0 +1,7 @@
+.device-registration-container {
+  width: 100%;
+}
```
SmitGala: remove

